PR #25, opened 2020-03-14
Added SD and .editorconfig

SteveWallace:
Updated with proper quoting and added basic .editorconfig to stop prettier from double quoting


scrapers.js
```diff
@@ -61,7 +63,8 @@ let scrapers = [
     }
   },
   {
-    url: 'https://opendata.arcgis.com/datasets/8840fd8ac1314f5188e6cf98b525321c_0.csv',
+    url:
+      'https://opendata.arcgis.com/datasets/8840fd8ac1314f5188e6cf98b525321c_0.csv',
```
lazd: Still dropping these guys down to the next line, can we prevent that?

SteveWallace: oh, sorry didn't notice that

Yeah, give me one minute.

scrapers.js
```diff
@@ -302,19 +320,35 @@ let scrapers = [
       let $ = await fetch.page(this.url);
 
       let cases = 0;
-      cases += parse.number($('p:contains("Number of PHL positives")').first().text().split(': ')[1]);
-      cases += parse.number($('p:contains("Number of commercial lab positives")').first().text().split(': ')[1]);
+      cases += parse.number(
+        $('p:contains("Number of PHL positives")')
+          .first()
+          .text()
+          .split(': ')[1]
```
lazd: Chaining is much clear like this, so let's keep it the way it is in your PR.

Can we make prettier do this automatically before commits?

scrapers.js
```diff
@@ -200,11 +213,10 @@ let scrapers = [
         if (string === 'None') {
           return 0;
         }
-        if (matches = string.match(/(\d+) of (\d+)/)) {
+        if ((matches = string.match(/(\d+) of (\d+)/))) {
           // Return the high number
           return parse.number(matches[2]);
-        }
-        else {
+        } else {
```
lazd: Definitely want `else` on a newline.

lazd:
Thanks so much for working through the weird coding style bits here. In the haste of setting this project up, I didn't bother with the usual, so getting this PR in with Prettier behaving right will definitely help ease some growing pains.

scrapers.js
SteveWallace: on saves, yes. not sure on commits? I'll verify

lazd:
Also, no need to re-send the PR once you get it back up to date with master, just force push to your branch.

SteveWallace:
@lazd updated. Turns out the prettier team is very opinionated about where an else/elseif goes. They won't add the option for it to be on a new line. Workaround is to add a stub comment between the bracket and else. I did that for the one that's in scraper.js.

lazd:
Excellent contribution. Thank you!

PR #27, opened 2020-03-14
Ut counties

osvyg:


lazd:
Excellent, thank you!

Note, I adjusted indentation; this repo uses two spaces per tab. I also changed double to single quotes and nitpicked some other spacing bits :)

PR #30, opened 2020-03-14
Ohio

SteveWallace:
Parsing a paragraph where they outline the counties with cases. I suspect this one will be pretty brittle, maybe they'll maintain the same formatting for a few days 💃


lazd:
Can you rebase? Thanks!

lazd:
Also looks like TN got mixed up in here.

PR #31, opened 2020-03-14
Fix network timeout causing hangs

qgolsteyn:
`Needle` by default will wait indefinitely when fetching if the host does not send any data back, or the data takes very long to be retrieved.

An example url causing issue is from the South Dakota website (https://doh.sd.gov/news/Coronavirus.aspx#SD) which does not seem to send any data back currently.

This PR adds default timeouts for opening connections (5s), waiting for a response (5s), and downloading data (30s).

Feel free to let me know if these defaults should be tweaked.

### DIFF
--- a/lib/fetch.js
+++ b/lib/fetch.js
@@ -8,7 +8,10 @@ import * as transform from './transform.js';
 // Spoof Chrome, just in case
 needle.defaults({
   parse_response: false,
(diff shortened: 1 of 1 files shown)


lazd:
This is awesome, I'll merge it as-is.

That said, what about retires? Let's discuss in https://github.com/lazd/coronadatascraper/issues/32

PR #34, opened 2020-03-14
added TN and OH

SteveWallace:


scrapers.js
```diff
@@ -1148,6 +1148,53 @@ let scrapers = [
...
+      let counties = [];
+      let $ = await fetch.page(this.url);
+      let $paragraph = $('p:contains("Number of counties with cases:")').text();
+      let regExp = /\(([^)]+)\)/;
+      let parsed = regExp.exec($paragraph);
+      let arrayOfCounties = parsed[1].split(',');
+      arrayOfCounties.map(county => {
+        let splitCounty = county.trim().split(' ');
+        counties.push({
+          county: splitCounty[0] + ' County',
```
lazd: throw a `parse.string(splitCounty[0])` in to clean out whitespace, whack random zero-width chars, etc.

scrapers.js
```diff
@@ -1148,6 +1148,53 @@ let scrapers = [
...
+          cases: splitCounty[1]
```
lazd: Throw a `parse.number()` around this to make it a real number.

SteveWallace: added

scrapers.js
SteveWallace: Added

lazd:
Nice work, thank you!

PR #38, opened 2020-03-15
Daily percent change

osvyg:
Resolves https://github.com/lazd/coronadatascraper/issues/19

### DIFF
--- a/index.js
+++ b/index.js
@@ -185,6 +185,8 @@ async function scrapeData() {
 async function writeData({ locations, featureCollection }) {
   let date = process.env['SCRAPE_DATE'] ?  '-' + process.env['SCRAPE_DATE'] : '';
 
+  locations = await addDailyPercentChange(locations);
+
   await fs.ensureDir('dist')
 
   await fs.writeFile(path.join('dist', `data${date}.json`), JSON.stringify(locations, null, 2));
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -207,4 +209,37 @@ async function generate(date, options = { findFeatures: true, findPopulations: t
...
+  return locations.map(location => {
+    let percentChange = "NA";
+    let locationDataYesterday = dataFromYesterday.find(d => d.county === location.county);
+    if (locationDataYesterday !== undefined) {
+      let casesDiff = location.cases - locationDataYesterday.cases;
+      if (casesDiff > 0) {
+        if (locationDataYesterday.cases > 0) {
+          percentChange = (casesDiff / locationDataYesterday.cases) * 100
+        } else {
+          percentChange = location.cases * 100
```
osvyg: I am not sure if this should be this or just 100

index.js
```diff
@@ -207,4 +209,37 @@ async function generate(date, options = { findFeatures: true, findPopulations: t
...
+  let filePath = './' + path.join('dist', `data-${yesterday}.json`);
+  let dataFromYesterday = [];
+  try {
+    dataFromYesterday = await fs.readJSON(filePath);
+  } catch (err) {
+    console.error('  ❌ JSON file "%s" not found percentages will not be calculated: ', filePath, err);
+    return locations
+  }
+  return locations.map(location => {
+    let percentChange = "NA";
```
osvyg: should this just be 0?

lazd:
So I'm actually working on doing this for the forthcoming timeseries data, where we know we have data from yesterday. Though your approach is valid, it assumes yesterday's data is present. I think rate change makes more sense in the context of timeseries.

Check out what I've done here: https://github.com/lazd/coronadatascraper/tree/timeseries

osvyg:
Sounds good

PR #39, opened 2020-03-15
First go at automating deployment

### LINKED ISSUE #5: Automatic daily deployment
Data should be deployed daily at 9PM PST.

qgolsteyn:
This PR adds a Github Action to deploy the website every day at 9PM PST. It builds the CSV data and site, and deploys it on Github Pages.

Closes #5 

**Known issues**
- Not linked to releases, but probably should. This could be added with this PR.
- Fails if we disable TLS checks


lazd:
Interesting that it fails with `--insecure-http-parser`... That's required to hit some of these sites since their certs are questionable.

And yeah JHU data breaks daily, so we just have to delay auto deployment until it's unbroken...

lazd:
Interesting that it fails with `--insecure-http-parser`... That's required to hit some of these sites since their certs are questionable.

And yeah JHU data breaks daily, so we just have to delay auto deployment until it's unbroken...

Great work so far!

qgolsteyn:
`--insecure-http-parser` works fine actually, so disabling TLS checks might be the only restriction imposed by Github Actions

qgolsteyn:
Happy to hear feedback on this PR

qgolsteyn:
I added an additional step to push the cache submodule following deployment. Can't test it fully as I don't have permission to push to that repo, but it should theoretically work.

lazd:
@qgolsteyn I added you to the cache repo, please test and let me know if this is good to go!

qgolsteyn:
Will do! Working on other things for the moment, but will get to it in a few hours

qgolsteyn:
Moved to using tags on master as the trigger mechanism. This would allow us to deploy when we create a new release.

PR #40, opened 2020-03-15
adding method to calculate active case

osvyg:
Resolves https://github.com/lazd/coronadatascraper/issues/33

### DIFF
--- a/index.js
+++ b/index.js
@@ -172,6 +172,7 @@ async function scrapeData() {
     else {
       counties++;
     }
+    location["active"] = transform.getActiveFromLocation(location)
   }
 
   console.log('✅ Data scraped!');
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -182,9 +182,21 @@ async function scrapeData() {
   return { locations };
 };
 
+function calculateActive(locations) {
+  return locations.map(l => {
```
lazd: Can we move this inner arrow function to `lib/transform.js` called `getActiveFromLocation`?

index.js
```diff
@@ -182,9 +182,21 @@ async function scrapeData() {
...
+    let recovered = l["recovered"] !== undefined ? l["recovered"] : 0;
+    l["active"] = cases - deaths - recovered;
+    return l;
+  });
+}
+
 async function writeData({ locations, featureCollection }) {
   let date = process.env['SCRAPE_DATE'] ?  '-' + process.env['SCRAPE_DATE'] : '';
 
+  locations = calculateActive(locations);
```
lazd: Let's do this above inside of `scrape()`, that way it's available further down the pipeline and to the API.

Maybe you can add it to the loop that's currently calculating counties/states/countries so we're not looping twice.

index.js
osvyg: fixed

index.js
osvyg: fixed

lazd:
Excellent work, thank you!

index.js
```diff
@@ -124,6 +124,7 @@ async function scrapeData() {
     else {
       counties++;
     }
+    location["active"] = transform.getActiveFromLocation(location)
```
lazd: @osvyg just realized this should probably check if `location.active` is already defined... perhaps the scraper or datasource already has it... Should we use that value if so?

osvyg: True, ill do a check. I'll create a new PR

osvyg: Yeah, I think we should trust what the active data is coming back from the scraper. If none then we calculate it

Issue #48, opened 2020-03-15
Use YYYY-MM-DD everywhere


lazd:
Currently, sometimes we have YYYY-M-D. Switch to YYYY-MM-DD for ease of sort and consistency.

Wikunia:
Maybe you can give a place where this can be done? Do you have a post-processing step?

lazd:
@Wikunia I believe @qgolsteyn is working on it. Thank you!

ryanblock:
I can look at this (I assume with @HerbCaudill).

ryanblock:
This is a fuzzily defined issue but we're on our way! See: https://github.com/covidatlas/coronadatascraper/commit/58ce17e3bd986bed82833fa1685a485684aa1d89

Issue #42, opened 2020-03-15
Replace hand-rolled date functions with moment.js or similar


lazd:
They're getting pretty gross. Let's use a library.

SteveWallace:
Do you want to just re factor the datetime.js to use moment or would you rather remove that helper and just use moment where we need it? Less change and probably better encapsulation with the former probably.

qgolsteyn:
See #50. We still need to update the codebase with the new `datetime.js` before the PR can be merged.

HerbCaudill:
Moment & Luxon etc. are great for what they do, but **they don't solve the problem we're having here**, which is that most of the time we don't care at all about times or time zones and we just need to work with stripped-down dates. 

I've been down this exact road before (hand-rolled date functions -> moment.js -> js-joda) with [my company's product](devresults.com). The happy place we ended up at was to eliminate the use of the native JS Date object throughout the codebase, and to use the appropriate js-joda classes instead as appropriate:

- [LocalDate](http://js-joda.github.io/js-joda/manual/LocalDate.html) for when you only care about calendar date, not time or time zone 👈 **99% of this project's use cases**
- [LocalTime](http://js-joda.github.io/js-joda/manual/LocalTime.html) for when you only care about time of day, not date, and not in any particular time zone
- [LocalDateTime](http://js-joda.github.io/js-joda/manual/LocalDateTime.html) for when you care about time of day + calendar date, but not time zone
- [ZonedDateTime](http://js-joda.github.io/js-joda/manual/ZonedDateTime.html) for when you care about time of day + calendar date in a specific time zone

Mixing these 4 very different combinations of concerns in a single object is a recipe for heartache and despair. 

js-joda is fast, solid and battle-tested: It replicates the API of [joda-time](https://github.com/JodaOrg/joda-time), which is the gold standard in the Java world, and replicates its test suite (1700+ tests). 

@lazd I can do this work tomorrow (Sunday, CET) if you give me the green light.

hyperknot:
@HerbCaudill we definitely need to care about timezones throughout this project. Soon timezones will be everywhere. We will run scrapers say hourly, and then want to transform the acquisition time to local time for every locale we are working with.

We definitely need a reliable and tested date library which handles timezones perfectly.

HerbCaudill:
@hyperknot For those cases, then, you can use joda's ZonedDateTime, which works beautifully. 

[...]

PR #50, opened 2020-03-15
Improve datetime functions

qgolsteyn:
This PR ensures all dates used are in the PST timezone, and adds associated tests to ensure the behaviour of these functions is correct.

Closes #50 

Would recommend testing it locally to make sure caching is still working as intended before merging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,5 +34,8 @@
     "@babel/preset-env": "^7.8.7",
     "babel-jest": "^25.1.0",
     "jest": "^25.1.0"
+  },
+  "engines": {
+    "node": ">=13.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "gh-pages": "^2.2.0",
     "jest-each": "^25.1.0",
     "moment": "^2.24.0",
+    "moment-timezone": "^0.5.28",
     "needle": "^2.3.3",
     "puppeteer": "^2.1.1",
     "yargs": "^15.3.0"
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "editorconfig": "^0.15.3",
     "gh-pages": "^2.2.0",
     "jest-each": "^25.1.0",
+    "moment": "^2.24.0",
     "needle": "^2.3.3",
     "puppeteer": "^2.1.1",
     "yargs": "^15.3.0"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,7 @@ name: Test CI
 on: push
 
 jobs:
-  build:
+  test:
     runs-on: [macOS-latest]
     steps:
       - uses: actions/checkout@v1 # without submodules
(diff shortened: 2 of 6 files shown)


qgolsteyn:
Hold off before merging, this is not fully integrated with `parser.js` and `timeseries.js`

lazd:
@qgolsteyn can we do #48 as part of this? For cache as well.

hyperknot:
I opened #194 and I see there is work already here. What I recommend:
- Use [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) standard, as detailed in #194. This is an industry standard human readable data function, with very well tested parsing and writing functions in all date libraries in all programming languages.
- We should not roll our own date methods.
- I recommend Luxon the recommended library from Moment.js developers. Think of it as a modern version of Moment.js.
- We should use UTC dates everywhere.

hyperknot:
Luxon has [.toISODate()](https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#instance-method-toISODate) and [.fromISO()](https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#static-method-fromISO), super well tested methods.

lazd:
> We should not roll our own date methods.
Yeah uhg, I should have started with moment or Luxon, but I was lazy.

ryanblock:
Would like to suggest [Spacetime](https://github.com/spencermountain/spacetime) as an alternative. Super lightweight and pretty flexible. Example impl: https://github.com/covidatlas/covidatlas/blob/master/src/shared/utils/get-date.js

package.json
```diff
@@ -34,5 +36,8 @@
     "@babel/preset-env": "^7.8.7",
     "babel-jest": "^25.1.0",
     "jest": "^25.1.0"
+  },
+  "engines": {
+    "node": ">=13.0.0"
```
ryanblock: Stable Node.js only, please!

lazd: Yeah I think that's leftover from before

HerbCaudill:
Please see my comment [here](https://github.com/lazd/coronadatascraper/issues/42#issuecomment-602115257). Moment and Luxon (and Spacetime AFAICT) all replicate the JS date object's original sin of forcing you to attach times and time zones to your dates, which makes everything much more complicated than it needs to be.

hyperknot:
@ryanblock I don't see what does Spacetime bring over luxon. Luxon is super stable and well tested. In Spacetime I don't even see the basic `fromISO()` and `toISO()` and `toISODate()` methods.

qgolsteyn:
#228 mentioned using [`js-joda`](https://js-joda.github.io/js-joda/manual/LocalDate.html). This library provides a `LocalDate`, which represents a date with no time and no time zone in the ISO-8601 calendar system. It does not use the Date object, does not force you to set a timezone. Could be an interesting option.

qgolsteyn:
Closing, this may not be the right approach. #42 for the up-to-date conversation.

PR #55, opened 2020-03-15
Update Santa Clara County

jbencina:
This change updates the pull for the Santa Clara County website. The page code dynamically populates the DOM from an embedded JSON array. This made a simple HTML parsing not possible. Instead, we find the script tag containing the JSON array, use regex to parse it, and then feed back the counts as normal.


lazd:
This is EXCELLENT. Thank you!

PR #56, opened 2020-03-15
Puppet fix and IA fix

SteveWallace:
changed the puppet function a bit to parse html and send it back over as cheerio object. Added user agent and view port as well. 

Corrected IA to use the new pattern. 

Had issues with getting my sub modules all synced up, I don't think it deleted anything though?

### DIFF
--- a/lib/fetch.js
+++ b/lib/fetch.js
@@ -112,10 +112,21 @@ function csv(url, date) {
 async function headless(url, date) {
   // todo: figure out how to read from cache
   const browser = await puppeteer.launch();
-  const page = await browser.newPage();
-  await page.goto(url);
-
-  return { browser, page };
+ const pPage = await browser.newPage();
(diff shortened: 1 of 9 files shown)


lib/fetch.js
```diff
@@ -112,10 +112,21 @@ function csv(url, date) {
...
+ await pPage.setViewport({width: 1280, height: 800, isMobile: false})
+ const response = await pPage.goto(url, 
+  {
+      timeout: 25000, 
+      waitUntil: 'networkidle2'
+  });
+  
+  if (response._status < 400) {
+      await pPage.waitFor(3000);
+      let html = await pPage.content();
```
lazd: This is going to need to be cached -- badly. Lets follow the same patterns for caching as used in `fetch`

lib/fetch.js
```diff
@@ -112,10 +112,21 @@ function csv(url, date) {
...
+      let page = await cheerio.load(html)
+      return {browser, page};
```
lazd: If we've loaded the page as HTML, I say go ahead and return that and close the browser immediately

PR #57, opened 2020-03-15
Update Ventura & Santa Cruz County

jbencina:
This PR updates Ventura and Santa Cruz County. Santa Cruz county changed its formatting from H1 to H2 tag object. Ventura County had an issue with JS not populating the count text, however the counts are available in the the data-from or data-to attributes.


lazd:
Fantastic contribution. Thank you!

PR #59, opened 2020-03-15
Map Congo (Kinshasa) to COD instead of COG

mark-otaris:
Congo-Kinshasa is the Democratic Republic of the Congo, which has three-letter code COD, while Congo is the different Republic of the Congo, which has three-letter code COG.


lazd:
Awesome. Good catch!

PR #60, opened 2020-03-15
Use HTTPS for cache submodule

mark-otaris:
This allows cloning with submodules without SSH. (`git submodule sync` will be needed to update the URL.)

Edit: This would conflict with updating the cache repository automatically using the submodule, but that seems best done with a separate SSH clone.


lazd:
Thanks! I had others get bit by this.

PR #62, opened 2020-03-15
Washington state headless fix

SteveWallace:
just swapped Washington over to the headless pull


SteveWallace:
Ohio changed url's so I updated that as well. Was going to start refactoring with moment.js and didn't realize I was in this branch <sigh>

SteveWallace:
killing this one since it's already fixed, will add another with just the OH change

PR #63, opened 2020-03-16
Ohio fix

SteveWallace:
updates Ohio url to new one

### DIFF
--- a/scrapers.js
+++ b/scrapers.js
@@ -1314,7 +1314,7 @@ let scrapers = [
   {
     state: 'OH',
     country: 'USA',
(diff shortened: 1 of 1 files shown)


lazd:
Seems like both URLs  work, any reason to switch?

Note that we'll need to use the old URL for dates < 2020-3-15. See other scrapers for examples on how do to that.

SteveWallace:
They changed the page between when I did this and now!

lazd:
@SteveWallace looks like this might be needed after all, Ohio stopped working and the old URL has a note that the info moved. Can you look into it?

SteveWallace:
fixed in other PR

PR #64, opened 2020-03-16
Add 7 new California counties

jbencina:
Adds 7 new counties to the California data:

- San Bernardino
- San Joaquin
- Merced
- Marin
- Butte
- Kings
- Mendocino


lazd:
Excellent work. Some of these will be brittle; can you note down which of these counties we need to reach out to to ask for them to provide a `<table>`?

PR #67, opened 2020-03-16
Maryland counties

SteveWallace:
added Maryland in with a check for Baltimore City so it doesn't add 'County' to it.

### DIFF
--- a/scrapers.js
+++ b/scrapers.js
@@ -1486,7 +1486,29 @@ let scrapers = [
       })
       return counties;
     }
-  }
+  },
+  {
+    state: 'MD',
+    country: 'USA',
+    url: 'https://coronavirus.maryland.gov/',
+    scraper: async function() {
+      let counties = [];
+      let $ = await fetch.headless(this.url);
+      let paragraph = $('p:contains("Number of Confirmed Cases:")').next('p').text();
+  
+      paragraph.split(')').map(splitCounty => {
+      if(splitCounty.length > 1){
+       let county = transform.addCounty(parse.string(splitCounty.substring(0, splitCounty.indexOf('(')).trim()))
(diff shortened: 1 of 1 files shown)


lazd:
AWESOME!

scrapers.js
```diff
@@ -1486,7 +1486,33 @@ let scrapers = [
...
+         county = transform.addCounty(county)
+       }
+       let cases = parse.number(splitCounty.substring(splitCounty.indexOf('(')+1, splitCounty.length).trim())
+       counties.push({
+         county,
+         cases
+       })}
+       
+      })
+      return counties;
```
lazd: Please add this, it will sum all of the counties and consider it the state total:
```suggestion
      counties.push(transform.sumData(counties));

      return counties;
```

We need to do this for all states we have 100% complete data for, that way we override JHU with our own, known-good data.

scrapers.js
```diff
@@ -1486,7 +1486,33 @@ let scrapers = [
...
+      let counties = [];
+      let $ = await fetch.headless(this.url);
+      let paragraph = $('p:contains("Number of Confirmed Cases:")').next('p').text();
+  
+      paragraph.split(')').map(splitCounty => {
+      if(splitCounty.length > 1){
+       let county = parse.string(splitCounty.substring(0, splitCounty.indexOf('(')).trim())
+       //check for Baltimore City
+       if (county !== 'Baltimore City') {
+         county = transform.addCounty(county)
```
lazd: If it's a city, you can treat it like one by setting `city`, and if you don't know the county, you can omit it.

lazd:
Turns out we can actually pull this from a CSV: http://opendata.arcgis.com/datasets/ca77764e722c442986ef6514da88411c_0.csv

PR #73, opened 2020-03-16
add nh.gov scraper

Smona:
Although the legends for the two "tested" rows are a bit confusing, i believe i picked the correct one.

Looks like this site is reporting more cases than the cdc is.

### DIFF
--- a/scrapers.js
+++ b/scrapers.js
@@ -24,6 +24,17 @@ import * as fs from './lib/fs.js';
 
 let scrapers = [
   {
+    state: 'NH',
+    country: 'USA',
+    url: 'https://www.nh.gov/covid19/',
+    scraper: async function() {
+      let $ = await fetch.page(this.url);
+      const cases = parse.number($('th:contains("Number of Persons with COVID-19") + td').html());
+      const tested = parse.number($('th:contains("Total Number of Persons Tested") + td').html());
+      return { cases, tested };
+    }
+  },
+  {
     state: 'AZ',
     country: 'USA',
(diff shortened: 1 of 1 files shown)


scrapers.js
```diff
@@ -23,6 +23,17 @@ import * as fs from './lib/fs.js';
...
 
 let scrapers = [
+  {
+    state: 'NH',
+    country: 'USA',
+    url: 'https://www.nh.gov/covid19/',
+    scraper: async function() {
+      let $ = await fetch.page(this.url);
+      const cases = parse.number($('th:contains("Number of Persons with COVID-19") + td').html());
+      const tested = parse.number($('th:contains("Total Number of Persons Tested") + td').html());
```
lazd: Use `.text()` in case they add formatting.

Smona:
Ah, I see. I didn't realize y'all were set on state-level data. Might be worth noting that in the README or the sources sheet for enthusiastic but uninformed would-be contributors like myself :)

lazd:
Yeah! We should really note that! The builds do give stats on duplicate data though, so at least you'll know if you add something that duplicates another data source.

lazd:
That said, we definitely appreciate the contribution and look forward to more! What can you pick from here to import? https://docs.google.com/spreadsheets/d/1T2cSvWvUvurnOuNFj2AMPGLpuR2yVs3-jdd_urfWU4c/edit#gid=0

Smona:
FWIW, I'm not seeing any warnings about duplicate data in the build output. Although I do see the CDC data now.

Glad to hear it! Just about to send you an email about the GA daily report page.

PR #75, opened 2020-03-16
change dates to YYYY-MM-DD

osvyg:
Resolves https://github.com/lazd/coronadatascraper/issues/48

I am not sure if this is all that it needs to do?

### DIFF
--- a/lib/fetch.js
+++ b/lib/fetch.js
@@ -80,7 +80,7 @@ async function fetch(url, type, date) {
       date = process.env['SCRAPE_DATE'];
     }
     else if (!date) {
-      date = datetime.getYYYYMD();
+      date = datetime.getYYYYMMDD();
     }
   }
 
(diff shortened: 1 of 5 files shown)


lib/fetch.js
```diff
@@ -80,7 +80,7 @@ async function fetch(url, type, date) {
       date = process.env['SCRAPE_DATE'];
     }
     else if (!date) {
-      date = datetime.getYYYYMD();
+      date = datetime.getYYYYMMDD();
```
mark-otaris: This date is used to retrieve from cache, which uses YYYY-M-D.

lazd: I'll go ahead and change the cache to use YYYY-MM-DD after this

lazd:
@osvyg perhaps we should do this after #42, perhaps as part of #50?

lib/fetch.js
ryanblock: Glad someone took a look at this, it was on my list!

qgolsteyn:
Closing as #50 will cover these changes

PR #83, opened 2020-03-16
Calculate ratings for datasources based on objective criteria

### LINKED ISSUE #82: Ratings for data sources
Scrapers should get a objective rating based on:

* ease of machine readability
* the completeness of available data
* timelines of updates
* consistency

We should advertise these ratings and actively point to them when reaching out to government department to demand better data.

lazd:
Closes #82 

This PR adds rating calculation based on available fields, data type, encryption, whether we have to access it headless, and whether it has time series data. The maximum rating is 1.

In addition, it uses the calculated rating to break ties when deduping data that has identical priorities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 # coronadatascraper
 > A scraper that pulls coronavirus case data from verified sources.
 
+This project exists to pull county-level data for COVID-19 from verified, high-quality sources.
+
+Every piece of data produced includes the URL where the data was sourced from as well as a rating of the source's technical quality (completeness, machine readability, best practices -- not accuracy).
+
 ## Where's the data?
 
 http://blog.lazd.net/coronadatascraper/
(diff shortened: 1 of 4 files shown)


lazd:
@jbencina thoughts?

jbencina:
I think this is a good first iteration. The values make intuitive sense and we're adding some objectivity to the process. We should just be careful to interpret this as the literal quality of how the data is pulled, vs. whether the source itself is credible or not.

qgolsteyn:
Looks great! I would suggest moving the ratings to a separate `.json` or `.js` file at the top of the project, so we can easily view/edit these ratings, and import them in the function.

lazd:
@qgolsteyn That's an excellent idea. They should probably be included in `report.json`!

qgolsteyn:
That would be nice! Maybe (although maybe another PR) include a list of scrappers that were dropped due to a better source being present?

#87 to track that thought

lazd:
@qgolsteyn agreed, let's report in another PR. Do you think this is ready to merge?

qgolsteyn:
Yes, LGTM. Go for it 🚀

PR #84, opened 2020-03-16
[minor] AUS Australia from health.gov.au

camjc:
Follows pattern in README.md for html scraping, happy to adjust, message me.

### DIFF
--- a/scrapers.js
+++ b/scrapers.js
@@ -2027,6 +2027,35 @@ let scrapers = [
       return counties;
     }
   },
+  {
+    country: 'AUS',
+    url: 'https://www.health.gov.au/news/health-alerts/novel-coronavirus-2019-ncov-health-alert/coronavirus-covid-19-current-situation-and-case-numbers',
+    type: 'table',
+    priority: 1,
+    scraper: async function() {
+      let states = [];
+      let $ = await fetch.page(this.url);
+
+      let $table = $('.health-table__responsive > table');
+
+      let $trs = $table.find('tbody > tr:not(:first-child):not(:last-child)');
+
+      $trs.each((index, tr) => {
+        let $tr = $(tr);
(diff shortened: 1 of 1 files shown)


scrapers.js
```diff
@@ -1686,6 +1686,30 @@ let scrapers = [
...
+        let $tr = $(tr);
+        let state = parse.string($tr.find('td:first-child').text());
+        let cases = parse.number($tr.find('td:nth-child(2)').text());
+        states.push({
+          state: state,
+          cases: cases
+        });
+      });
+
+      return states;
```
lazd: This little ditty makes your scraper also include data for `AUS` as a whole:

```suggestion
      // Add data for AUS itself
      states.push(transform.sumData(states));

      return states;
```

scrapers.js
```diff
@@ -1686,6 +1686,30 @@ let scrapers = [
       };
     }
   },
+  {
+    country: 'AUS',
+    url: 'https://www.health.gov.au/news/health-alerts/novel-coronavirus-2019-ncov-health-alert/coronavirus-covid-19-current-situation-and-case-numbers',
```
lazd: This little number makes your data authoritative vs other sources for AUS and its states, as well as provides the type so our forthcoming rating system #82 can rank the data source.

```suggestion
    url: 'https://www.health.gov.au/news/health-alerts/novel-coronavirus-2019-ncov-health-alert/coronavirus-covid-19-current-situation-and-case-numbers',
    type: 'table',
    priority: 1,
```

camjc:
@lazd Cheers for the CR, let me know if there's anything else :)

P.S. I wonder if theres a different modularisation that would reduce the need for everyone to do rebases (glob require-ing a file for each country from a folder or something). NaBD but something we can think about.

PR #93, opened 2020-03-17
updated and fixed OH, MD and LA

SteveWallace:
updated OH url, re-added code for MD and removed a dupe of LA that was causing issues.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -8,9 +8,6 @@ function number(string) {
   if (string === '') {
     return 0;
   }
-  if (string === undefined) {
-    return 0;
-  }
   return parseInt(string.replace(/[^\d\-]/g, ''), 10);
 }
 
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -8,6 +8,9 @@ function number(string) {
   if (string === '') {
     return 0;
   }
+  if (string === undefined) {
+    return 0;
+  }
   return parseInt(string.replace(/[^\d\-]/g, ''), 10);
 }
 
@@ -25,12 +28,15 @@ function float(string) {
   Remove line breaks, double spaces, zero-width space, asterisk, and trim the provided string
 */
 function string(string) {
+  if (string) {
   return string
     .replace(/\n/g, ' ')
     .replace(/\s+/, ' ')
     .replace(/\u200B/g, '')
     .replace(/\*/g, '')
     .trim();
+  }
+   else { return null }
 }
 
 export { number, float, string };
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -28,15 +28,12 @@ function float(string) {
   Remove line breaks, double spaces, zero-width space, asterisk, and trim the provided string
 */
 function string(string) {
-  if (string) {
   return string
     .replace(/\n/g, ' ')
     .replace(/\s+/, ' ')
     .replace(/\u200B/g, '')
     .replace(/\*/g, '')
     .trim();
-  }
-   else { return null }
 }
 
 export { number, float, string };
--- a/scrapers.js
+++ b/scrapers.js
@@ -1692,13 +1692,24 @@ let scrapers = [
   {
     state: 'OH',
     country: 'USA',
-    url: 'https://coronavirus.ohio.gov/wps/portal/gov/covid-19/',
     scraper: async function() {
-      let counties = [];
-      let $ = await fetch.page(this.url);
-      let $paragraph = $('p:contains("Number of counties with cases:")').text();
-      let parsed = $paragraph.replace(/([()])/g, '').replace('* Number of counties with cases: ','');
-      let arrayOfCounties = parsed.split(',');
+      let counties = []
+      let arrayOfCounties  = []
+      if (datetime.scrapeDateIsBefore('2020-3-16')) {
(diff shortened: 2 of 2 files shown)


SteveWallace:
Oh and added a little checker for undefined strings on the parser.

lib/parse.js
```diff
@@ -25,12 +28,15 @@ function float(string) {
   Remove line breaks, double spaces, zero-width space, asterisk, and trim the provided string
 */
 function string(string) {
+  if (string) {
```
qgolsteyn: JS being JS, the string `""` is actually falsy, which is why the test failed...

```suggestion
  if (string === undefined) {
```

SteveWallace: That check is for a string not for undefined. @lazd could reverse the logic I guess and look for undefined instead?

SteveWallace: oh, nevermind! that was code I added and meant to delete

lazd: Yeah in this case, if it's an empty string or not truthy, we want to return `null` so it's discarded. This isn't for parsing numbers, so `'0'` won't be an issue (yes, `'0' == false`)

scrapers.js
```diff
@@ -1692,14 +1692,13 @@ let scrapers = [
   {
     state: 'OH',
     country: 'USA',
-    url: 'https://odh.ohio.gov/wps/portal/gov/odh/know-our-programs/Novel-Coronavirus/welcome/',
+    url: 'https://coronavirus.ohio.gov/wps/portal/gov/covid-19/',
     scraper: async function() {
       let counties = [];
       let $ = await fetch.page(this.url);
```
lazd:
```suggestion
      if (datetime.scrapeDateIsBefore('2020-3-16')) {
        this.url = 'https://odh.ohio.gov/wps/portal/gov/odh/know-our-programs/Novel-Coronavirus/welcome/';
      } else {
        this.url = 'https://coronavirus.ohio.gov/wps/portal/gov/covid-19/';
      }
      let $ = await fetch.page(this.url);
```

scrapers.js
```diff
@@ -1692,14 +1692,13 @@ let scrapers = [
   {
     state: 'OH',
     country: 'USA',
-    url: 'https://odh.ohio.gov/wps/portal/gov/odh/know-our-programs/Novel-Coronavirus/welcome/',
+    url: 'https://coronavirus.ohio.gov/wps/portal/gov/covid-19/',
```
lazd:
```suggestion
```
Remove the URL from here

scrapers.js
```diff
@@ -1880,6 +1841,32 @@ let scrapers = [
...
+       let cases = parse.number(splitCounty.substring(splitCounty.indexOf('(')+1, splitCounty.length).trim())
+       counties.push({
+         county,
+         cases
+       })}
+       
+      })
+      return counties;
+    }
+  },
```
lazd: Sorry for the merge issues! Thanks for sorting it out. Did you need to change `'LA'` above?

lib/parse.js
```diff
@@ -8,6 +8,9 @@ function number(string) {
   if (string === '') {
     return 0;
   }
+  if (string === undefined) {
+    return 0;
+  }
```
lazd: So I'm not sure about this. I feel undefined numbers are an indication of failure -- the scraper found nothing, so it should stop and we should report an error. @qgolsteyn thoughts?

scrapers.js
lazd: Ah, I see, it's present further up. Looking right!

scrapers.js
```diff
@@ -1692,14 +1692,13 @@ let scrapers = [
...
+    url: 'https://coronavirus.ohio.gov/wps/portal/gov/covid-19/',
     scraper: async function() {
       let counties = [];
       let $ = await fetch.page(this.url);
       let $paragraph = $('p:contains("Number of counties with cases:")').text();
-      let regExp = /\(([^)]+)\)/;
-      let parsed = regExp.exec($paragraph);
-      let arrayOfCounties = parsed[1].split(',');
+      let parsed = $paragraph.replace(/([()])/g, '').replace('* Number of counties with cases: ','');
+      let arrayOfCounties = parsed.split(',');
```
lazd: Is this backwards compatible with yesterday?

lib/parse.js
SteveWallace: actually that's exactly what it was. I added it when debugging OH. We can remove it, you're not wrong.

scrapers.js
SteveWallace: yeah, tested against 3/15 scrapes. works fine.

SteveWallace: sigh, nevermind. On wrong branch.

scrapers.js
lazd: Now that you're setting the URL below, you can accept this suggestion and remove it from above.

scrapers.js
lazd: Thanks for dealing with it... I know it can be annoying to make the scrapers backwards compatible, but it's gonna pay off when we have timeseries data for all the things

SteveWallace: np, updated the PR. Now does the proper parsing. However 3/15 is broken, we don't seem to have a cache for Ohio? 3/14 and 3/16 work with the new code.

lazd: Yes, we probably don't have 3/14. I will get it with the wayback machine once that work is compelte.

PR #94, opened 2020-03-17
Fix Glenn County to new url

### LINKED ISSUE #92: Glenn County, CA scraper broken
New URL is https://www.countyofglenn.net/dept/health-human-services/public-health/covid-19 and scraper might need to be updated.

Remember: preserve old scraper by hiding it behind `datetime.scrapeDateIsBefore('2020-3-16')` and adjust URLs accordingly.

@jbencina can you investigate?

jbencina:
Fixes #92 

Despite a new URL, the format is still the same and no changes were necessary to the parsing logic. Added inline comment to note the fragility due to unstructured format.


scrapers.js
```diff
@@ -1424,10 +1424,13 @@ let scrapers = [
     county: 'Glenn County',
     state: 'CA',
     country: 'USA',
-    url: 'https://www.countyofglenn.net/dept/health-human-services/public-health/welcome',
+    url: 'https://www.countyofglenn.net/dept/health-human-services/public-health/covid-19',
     scraper: async function() {
       let $ = await fetch.page(this.url);
```
lazd:
```suggestion
      if (datetime.scrapeDateIsBefore('2020-3-16')) {
        this.url = 'https://www.countyofglenn.net/dept/health-human-services/public-health/welcome';
      }
      else {
        this.url = 'https://www.countyofglenn.net/dept/health-human-services/public-health/covid-19';
      }

      let $ = await fetch.page(this.url);
```

For backwards compatibility

scrapers.js
```diff
@@ -1424,10 +1424,13 @@ let scrapers = [
     county: 'Glenn County',
     state: 'CA',
     country: 'USA',
-    url: 'https://www.countyofglenn.net/dept/health-human-services/public-health/welcome',
+    url: 'https://www.countyofglenn.net/dept/health-human-services/public-health/covid-19',
```
lazd:
```suggestion
```
No need now that it's set below

lazd:
Thanks! Note the changes for backwards compatibility; it's weird, but we're writing scrapers that can scrape a snapshot of the page at any given point in time, so we have to be able to scrape the old stuff too.

PR #95, opened 2020-03-17
Added Mono County, CA

jbencina:
This adds Mono County using the headless scraper. Checked that data is correctly output to json file.


lazd:
Thank you!

PR #99, opened 2020-03-17
ITA scraper for new italian data

enrichman:


lazd:
This is so much bigger than this PR shows. Thank you for the effort on the GeoJSON and population information over in https://github.com/lazd/coronavirus-data-sources/pull/5 -- your knowledge saved me hours :)

PR #104, opened 2020-03-17
[patch] add linting, and remove lint

camjc:
Please focus CR effort on commit ` [patch] manually applied linting fixes, CR carefully `. The rest are unlikely to break anything.

I've tried to reach a balance between consistency to airbnb style, and things that seem to be part of the original author's style choices. Happy to take your lead on this @lazd, but in any case this PR will go stale super quick so keen to get it in ASAP (and while we only have 3 other PRs open, so we don't force many people to rebase).

Hopefully this will make some of the code more readable, by consistency and cleaning up things like stray unused variables.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,17 @@
+{
+  "extends": ["airbnb-base", "plugin:jest/recommended", "plugin:jest/style", "plugin:prettier/recommended"],
+  "parser": "babel-eslint",
+  "rules": {
+    "jest/no-standalone-expect": "warn",
+    "no-async-promise-executor": "warn",
+    "no-await-in-loop": "warn",
+    "no-console": "off",
+    "no-continue": "warn",
+    "no-labels": "warn",
+    "no-param-reassign": "warn",
+    "no-plusplus": "off",
+    "no-restricted-syntax": "off",
+    "no-shadow": "warn",
+    "no-underscore-dangle": "off"
+  }
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,6 +12,7 @@
     "no-plusplus": "off",
     "no-restricted-syntax": "off",
     "no-shadow": "warn",
+    "import/extensions": ["error", "ignorePackages"],
     "no-underscore-dangle": "off"
   }
 }
--- a/cli.js
+++ b/cli.js
@@ -1,7 +1,4 @@
-import path from 'path';
-import generate from './index.js';
-import * as fs from './lib/fs.js';
-import * as stringify from './lib/stringify.js';
-import argv from './lib/cliArgs.js';
+import generate from './index';
+import argv from './lib/cliArgs';
 
 generate(argv.date, argv);
--- a/cli.js
+++ b/cli.js
@@ -1,4 +1,4 @@
-import generate from './index';
-import argv from './lib/cliArgs';
+import generate from './index.js';
+import argv from './lib/cliArgs.js';
 
 generate(argv.date, argv);
--- a/package.json
+++ b/package.json
@@ -48,7 +48,21 @@
     "eslint-plugin-import": "^2.20.1",
     "eslint-plugin-jest": "^23.8.2",
     "eslint-plugin-prettier": "^3.1.2",
+    "husky": "^4.2.3",
     "jest": "^25.1.0",
+    "lint-staged": "^10.0.8",
     "prettier": "^1.19.1"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged",
+      "pre-push": "npm run lint"
+    }
+  },
+  "lint-staged": {
+    "*.js": [
+      "eslint --fix",
+      "git add"
+    ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "@adobe/spectrum-css": "^2.18.0",
     "@turf/turf": "^5.1.6",
     "cheerio": "^1.0.0-rc.3",
-    "csv": "^5.3.2",
     "csv-parse": "^4.8.8",
     "csv-stringify": "^5.3.6",
     "editorconfig": "^0.15.3",
--- a/package.json
+++ b/package.json
@@ -1,30 +1,33 @@
 {
   "name": "coronadatascraper",
   "version": "1.0.0",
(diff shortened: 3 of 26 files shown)


cli.js
```diff
@@ -1,7 +1,4 @@
-import path from 'path';
-import generate from './index.js';
-import * as fs from './lib/fs.js';
-import * as stringify from './lib/stringify.js';
-import argv from './lib/cliArgs.js';
+import generate from './index';
```
qgolsteyn: Unfortunately, removing the `.js` extension breaks the imports 😢 We are using ES modules rather than CommonJS modules, meaning the extension needs to be specified in the import path.

package.json
```diff
@@ -1,43 +1,68 @@
 {
   "name": "coronadatascraper",
   "version": "1.0.0",
-  "description": "A scraper for Cornavirus data",
+  "description": "A scraper for Coronavirus data",
```
lazd: lol of course I did

package.json
```diff
@@ -1,43 +1,68 @@
...
+    "updateModules": "git submodule update --remote",
+    "winStart": "set NODE_OPTIONS='--insecure-http-parser' && node cli.js"
   },
   "dependencies": {
     "@adobe/spectrum-css": "^2.18.0",
     "@turf/turf": "^5.1.6",
     "cheerio": "^1.0.0-rc.3",
     "csv": "^5.3.2",
+    "csv-parse": "^4.8.8",
+    "csv-stringify": "^5.3.6",
```
lazd: We can remove `csv` itself if we add these as deps directly

cli.js
lazd: I'd say add extensions back is the right way to go, that was there are no surprises

package.json
camjc: IDM, which is better?

lazd: I say drop `csv`, thanks!

.eslintrc.json
```diff
@@ -0,0 +1,18 @@
...
+    "no-async-promise-executor": "warn",
+    "no-await-in-loop": "warn",
+    "no-console": "off",
+    "no-continue": "warn",
+    "no-labels": "warn",
+    "no-param-reassign": "warn",
+    "no-plusplus": "off",
+    "no-restricted-syntax": "off",
+    "no-shadow": "warn",
+    "import/extensions": ["error", "ignorePackages"],
```
camjc: Enforce .js, except for node modules

tasks/findFeatures.js
```diff
@@ -91,16 +97,9 @@ const generateFeatures = ({ locations, report, options }) => {
...
-  let featureCollection = {
-    type: 'FeatureCollection',
-    features: []
-  };
-
-  return new Promise(async (resolve, reject) => {
+  return new Promise(async resolve => {
     console.log('⏳ Generating features...');
 
-    const usStates = await fs.readJSON('./coronavirus-data-sources/lib/us-states.json');
```
osvyg: was this intentional?

tasks/findFeatures.js
```diff
@@ -174,33 +171,15 @@ const generateFeatures = ({ locations, report, options }) => {
...
-              ) ||
-              location.county && (
-                location.county === feature.properties.name ||
-                location.county === feature.properties.name_en ||
-                location.county === feature.properties.region
-              )
-            )
-          ) {
+        for (const feature of provinceData.features) {
+          if ((location.country === feature.properties.gu_a3 || location.country === feature.properties.adm0_a3) && ((location.state && (location.state === feature.properties.name || location.state === feature.properties.name_en || location.state === feature.properties.region)) || (location.county && (location.county === feature.properties.name || location.county === feature.properties.name_en || location.county === feature.properties.region)))) {
```
lazd: This is totally unreadable now, but TBH, no expression should look like this anyway. I'll split it out into variables and test those in the expression instead of slapping it all in the `if`.

tasks/findFeatures.js
camjc: Yep, its not used anywhere. Can easily chuck it back in once its referenced.

timeseries.js
```diff
@@ -1,26 +1,21 @@
-import generate from './index.js';
 import path from 'path';
+import generate from './index.js';
```
lazd: This is a nit I can get behind

timeseries.js
```diff
@@ -82,54 +67,39 @@ async function generateTidyCSV(timeseriesByLocation) {
...
+    for (const date of dates) {
+      const dateRow = row.slice();
       let hasData = false;
-      for (let type of caseDataProps) {
+      for (const type of caseDataProps) {
         if (location.dates[date] && location.dates[date][type]) {
           hasData = true;
         }
-        dateRow.push(location.dates[date] && location.dates[date][type] || '');
+        dateRow.push((location.dates[date] && location.dates[date][type]) || '');
```
lazd: Good catch!

tools/mapJHUCounties.js
```diff
@@ -26,17 +23,16 @@ async function findNameMapping() {
...
-      for (let feature of usCountyData.features) {
-        if (point && feature.geometry) {
-          let poly = turf.feature(feature.geometry);
+      for (const feature of usCountyData.features) {
+        if (feature.geometry) {
+          const poly = turf.feature(feature.geometry);
           if (turf.booleanPointInPolygon(point, poly)) {
             countyMap[location[STATE]] = feature.properties.name;
-            continue locationLoop;
+            continue;
```
lazd: Erm, we want to continue the outer loop here; not the loop over `usCountyData`.

tasks/findFeatures.js
camjc: Yeah agreed, feel like the original code for this is a little unreadable, if anything this highlights it as an opportunity ;)

timeseries.js
camjc: autofixed :)

osvyg:
LGTM 👍

tools/mapJHUCounties.js
camjc: Oh, gotcha. Yeah I never code in `continue` style, fixing now.

tasks/findFeatures.js
lazd: Ah yes, leftovers.

PR #107, opened 2020-03-17
correct 3166 codes for Kosovo and Palestine

### LINKED ISSUE #78: a few more ISO3166 names
it seems like Congo (Brazzaville) and Hong Kong have ISO3166 codes (COG and HKG).  (at least according to wikipedia).  i think that would just leave Kosovo without a code.

### LINKED ISSUE #88: Palestine and Kosovo codes
Palestine in ISO 3166 is defined as "PSE" (currently the code uses "PSR").

also, at least the World Bank seems to define Kosovo as "XKX", so maybe that would be a good code to use there.

greg-minshall:
XKX is from World Bank usage

Closes #88 
Closes #78

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -11,7 +11,7 @@ let countryCodes = JSON.parse(fs.readFileSync(path.join('coronavirus-data-source
   Override some incorrect country names
 */
 let countryMap = {
-  'Kosovo': 'KOS',
+  'Kosovo': 'XKX',
   'Congo (Kinshasa)': 'Congo, Democratic Republic of the',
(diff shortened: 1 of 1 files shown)


lib/transform.js
```diff
@@ -10,8 +10,8 @@ const countryCodes = JSON.parse(fs.readFileSync(path.join('coronavirus-data-sour
 /*
   Override some incorrect country names
 */
-const countryMap = {
-  Kosovo: 'KOS',
+let countryMap = {
```
lazd:
```suggestion
const countryMap = {
```

PR #120, opened 2020-03-18
Fix Glenn, Yolo, and San Joaquin CA counties

jbencina:
This fixes three CA counties.

- Yolo County started adding ** which broke the regex
- Glenn County changed from FONT to SPAN tags
- San Joaquin started using a table

### DIFF
--- a/scrapers.js
+++ b/scrapers.js
@@ -1209,10 +1209,21 @@ const scrapers = [
     async scraper() {
       const $ = await fetch.page(this.url);
 
-      // this is brittle as all hell
-      const $h3 = $('h3:contains("confirmed case")');
+      // Site currently contains information in a headline only
 
-      const matches = $h3.text().match(/there are (\d+) confirmed cases? in Yolo/);
+      if (datetime.scrapeDateIsBefore('2020-03-17')) {
+        const $h3 = $('h3:contains("confirmed case")');
+        const matches = $h3.text().match(/there are (\d+) confirmed cases? in Yolo/);
+
+        return {
+          cases: parse.number(matches[1])
+        };
+      }
+
+      // They started adding in asterisks like - there are **4 confirmed cases
+      const $h3 = $('h3:contains("confirmed case")');
+      const matches = $h3.text().match(/(\d+) confirmed case/);
+
       return {
         cases: parse.number(matches[1])
       };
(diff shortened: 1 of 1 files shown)


scrapers.js
```diff
@@ -1209,10 +1209,21 @@ const scrapers = [
...
+      if (datetime.scrapeDateIsBefore('2020-03-17')) {
+        const $h3 = $('h3:contains("confirmed case")');
+        const matches = $h3.text().match(/there are (\d+) confirmed cases? in Yolo/);
+
+        return {
+          cases: parse.number(matches[1])
+        };
+      }
+
+      // They started adding in asterisks like - there are **4 confirmed cases
```
lazd: lol thanks, YOLO

scrapers.js
```diff
@@ -1548,13 +1589,24 @@ const scrapers = [
...
+            .match(/Cases: (\d+)/)[1]
+        );
+        return {
+          cases
+        };
+      }
+
       const cases = parse.number(
-        $('font:contains("Glenn County COVID-19 Cases")')
+        $('span:contains("Glenn County COVID-19 Cases")')
```
lazd: You could put a conditional in the selector if you wanted:

```suggestion
        let tagName = datetime.scrapeDateIsBefore('2020-3-17') ? 'font' : 'span';
        $(tagName + ':contains("Glenn County COVID-19 Cases")')
```

The other options is to just do  `$(':contains("Glenn County COVID-19 Cases")').last()` -- but it's probably slower... Thoughts?

scrapers.js
```diff
@@ -1209,10 +1209,21 @@ const scrapers = [
     async scraper() {
       const $ = await fetch.page(this.url);
 
-      // this is brittle as all hell
+      // Site currently contains information in a headline only
+
+      if (datetime.scrapeDateIsBefore('2020-03-17')) {
+        const $h3 = $('h3:contains("confirmed case")');
+        const matches = $h3.text().match(/there are (\d+) confirmed cases? in Yolo/);
```
lazd: Ok curious: should we preserve the entire scraper and duplicate code behind the conditional? Will that make it easier to maintain in the future? Probably... So ignore my other comment about changing the selector with a ternary operator...

PR #123, opened 2020-03-18
Let scraper be an object keyed on date

### LINKED ISSUE #122: Let scraper be an object keyed by scrapeDate
```
scraper: {
  '2020-3-15': func1,
  '2020-3-16': func2,
  '2020-3-20': func3
}
```

It should match the closest date that is not greater than the current date.

lazd:
Closes #122

### DIFF
--- a/lib/datetime.js
+++ b/lib/datetime.js
@@ -85,6 +85,18 @@ function dateIsBefore(a, b) {
 }
 
 /*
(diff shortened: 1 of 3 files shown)


lazd:
@jbencina @qgolsteyn if this is good, let's merge so I can fix MD with it!

PR #124, opened 2020-03-18
adding colorado fix

osvyg:


lazd:
This works for timeseries too?
Was it just backwards?

PR #135, opened 2020-03-19
docs: very super ultra basic choropleth map for the current date

lazd:
It looks like this:
image

It uses a rather dubious algorithm to set the color. I fully expect someone to suggest something better. The color scheme isn't that great either.

But it's a map.

Grey: we have case data, but no population data OR "this location is not granular enough"
White: we have case data, and there are no cases


lazd:
I'm gonna go ahead and merge this without review on account of the fact that we'll definitely change it soon.

PR #139, opened 2020-03-19
Enables project to be run on non-experimental Node.js version via esm; removes SSL workaround for env var

### LINKED ISSUE #142: Node 12.x support
Node 12.x is the stable target we should be running in production.

ryanblock:
Closes #142


scrapers.js
```diff
@@ -818,6 +818,8 @@ const scrapers = [
     aggregate: 'county',
     ssl: false, // Error: unable to verify the first certificate
     async scraper() {
```
lazd: What if when we set `ssl: false` on a scraper whose scheme is https, it automatically sets `process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';` when starting and does `delete process.env.NODE_TLS_REJECT_UNAUTHORIZED;` after the request completes (or errors)?

Or, what if we just leave this on at all times? I guess then we won't see failures if/when a scraper has bad SSL certs, but maybe that's for the better?

cli.js
```diff
@@ -1,4 +1,6 @@
-import generate from './index.js';
-import argv from './lib/cliArgs.js';
+const imports = require('esm')(module);
+
+const generate = imports('./index.js').default;
+const argv = imports('./lib/cliArgs.js').default;
```
lazd: Can these be changed to normal `import`, or does it have to use `imports`?

package.json
```diff
@@ -18,13 +15,12 @@
     "deploy": "npm run dateData && npm run deploySite",
     "deploySite": "npm run buildSite && gh-pages -a -d dist/ -e .",
     "lint": "eslint .",
-    "postinstall": "node initSSL.js",
-    "start": "NODE_OPTIONS='--insecure-http-parser' node cli.js",
+    "start": "node cli.js",
     "test": "jest",
     "timeseries": "node timeseries.js",
```
lazd: What's curious is that we never passed that option to `timerseries` -- looks like a bug.

cli.js
ryanblock: Would be a duplicate declaration, and afaik not serve any purpose since this entry is only doing one thing. Surface area for changing back to completely pure es modules in the future is pretty minimal though!

scrapers.js
ryanblock: I think to keep things maximally secure we should not leave `NODE_TLS_REJECT_UNAUTHORIZED` on at all times, but will add a new setting that will flip that switch on and off programatically. Stand by!

qgolsteyn:
We moved away from setting `process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"` because GitHub Action appeared to block any requests to go through. I am working on some automated tests for fetching data, which could help us experiment with `process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"` to see if there is maybe a way to get around it.

ryanblock:
@qgolsteyn thanks! What specifically was GitHub Actions doing? Is it somehow proxying requests or something? I don't think it should be handling or munging HTTP in any way...

qgolsteyn:
I was wrong, `process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"` might actually work after all. #155 adds the ability to disable SSL in `fetch.js` rather than in the scraper task. It also comes with a test to see if this approach works in Github Actions. The test passed, so I have no problem with disabling SSL when necessary.

package.json
lazd: Fixed in master anyway, stoked to nuke it completely :)

package.json
```diff
@@ -18,12 +15,11 @@
...
     "deploySite": "npm run buildSite && gh-pages -a -d dist/ -e .",
     "lint": "eslint .",
-    "start": "NODE_OPTIONS='--insecure-http-parser' node cli.js",
+    "start": "node cli.js",
     "test": "jest",
     "timeseries": "NODE_OPTIONS='--insecure-http-parser' node timeseries.js",
     "update": "npm run updateModules && rm -rf cache/* && npm run start",
-    "updateModules": "git submodule update --remote",
-    "winStart": "set NODE_OPTIONS='--insecure-http-parser' && node cli.js"
+    "updateModules": "git submodule update --remote"
```
lazd: GOLD. Does this need to be `--recursive` though?

tasks/scrapeData.js
```diff
@@ -133,9 +133,17 @@ function addData(cases, location, result) {
...
+  if (rejectUnauthorized) {
+    // Important: this prevents SSL from failing
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+  }
   if (typeof location.scraper === 'function') {
     return location.scraper();
   }
+  if (rejectUnauthorized) {
+    delete process.env.NODE_TLS_REJECT_UNAUTHORIZED;
+  }
```
lazd: @qgolsteyn I prefer this approach -- that way we have the information surfaced for ratings. Thoughts?

lazd: That said, it's not as easily testable...

qgolsteyn: I feel both can exist? Although I could also modify the test to set `process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"` before it runs as well, so testing is a non-issue.

lazd: Yeah, it's fine if both exist...

qgolsteyn:
Hmm, we had to revert this PR as it generated this error:
```
file:///Users/samir/Documents/coronadatascraper/cli.js:1
const imports = require('esm')(module);
                ^
ReferenceError: require is not defined
    at file:///Users/samir/Documents/coronadatascraper/cli.js:1:17
    at ModuleJob.run (internal/modules/esm/module_job.js:109:37)
    at async Loader.import (internal/modules/esm/loader.js:141:24)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

PR #147, opened 2020-03-19
Updated Mono County, CA to include tests

jbencina:
This fixes the Mono County to correctly pull the right cases. It was reporting 0 however the number of cases is also 0 so no change to the data. I added the number of tests given. This was not available previously and is how I recognized the issue.


qgolsteyn:
Thanks for this PR!

PR #150, opened 2020-03-19
feat: add county to de fl pa

jeffchenoweth:
Latest updates to DE, FL and PA removed the 'County' transform


lazd:
Ya can't go wrong with `transform.addCounty()`. Thanks, @jeffchenoweth!

PR #155, opened 2020-03-19
test: ability to disable ssl

qgolsteyn:
This PR adds the ability to disable SSL checks when fetching new resources. It comes with a test to verify that this feature works in CI.


qgolsteyn:
Disabling SSL seems to be working in Github Actions. Not sure what error I faced a couple days ago, but that means we can take that approach!

lazd:
LGTM, let's merge this and rebase #139, then merge that!

PR #157, opened 2020-03-19
fix #145

### LINKED ISSUE #145: Tanzania has no country code
I just came across that Tanzania does not have it's country code in the `timeseries.csv`.
It should be `TZA` but instead showing `Tanzania`.

Wikunia:
Not sure what the desired outcome for Faroer Islands is so maybe this currently makes it worse.


qgolsteyn:
I think moving to ISO codes is the correct direction, so I will merge this. Thanks for this contribution!

PR #158, opened 2020-03-19
Link to Jupyter quickstart notebook

thatneat:
This notebook contains code to allow an analyst to get off the ground quickly with this data, as well as some example plots.


lazd:
Excellent work, @thatneat !

lazd:
@thatneat looks like the `corona data scraper` link needs to have `https://` in front of it, also the text should be `Corona Data Scraper`. Thanks!

thatneat:
Thanks! fixed.

PR #163, opened 2020-03-19
Enforce Node.js 12.x + fix commonjs type jic

ryanblock:


lazd:
Ok awesome, but now we need to add back `esm` and the other changes that got rolled back when the commit was reverted.

ryanblock:
Will have to kill this PR I think.

PR #175, opened 2020-03-20
feat: add tests for scrapers

qgolsteyn:
This PR adds the ability to test modified scrapers. Locally, it tests the scrapers changed from the last commit. In a PR, it tests scrapers changed compared to the master branch.

Currently, this test makes sure the scraper does not throw any exception. This could be expanded in the future.

### DIFF
--- a/scrapers/FRA/index.js
+++ b/scrapers/FRA/index.js
@@ -14,6 +14,8 @@ const scraper = {
   priority: 1,
   aggregate: 'state',
   async scraper() {
+    // make it fail
+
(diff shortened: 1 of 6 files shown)


scrapers/FRA/index.js
```diff
@@ -43,7 +45,8 @@ const scraper = {
       }
     }
     states.push(transform.sumData(states));
-    return states;
```
lazd: wait, poor FRA

qgolsteyn: Yeah, just making sure that my code works as intended

PR #188, opened 2020-03-20
filter timeseries using handsontable

Wikunia:
Makes it easier to see all data for a specific country or see countries where there are tests.
Filters


lazd:
Uh, YES! I had no idea we could do this.

Can you send this against the `rollup` branch? We are refactoring how the site is built.

Issue #79, opened 2020-03-16
Scraper for DEU (Germany) regions


lazd:
We need regional data for DEU.

hyperknot:
datawrapper.de writes in: https://blog.datawrapper.de/coronaviruscharts/#considerations

<p>For the <a href="#map-germany">state map of Germany</a>, we use numbers from the <a href="https://www.rki.de/DE/Content/InfAZ/N/Neuartiges_Coronavirus/Fallzahlen.html" target="_blank" rel="noopener"><strong>Robert Koch Insitute</strong></a>. This source is official, but updated slowly in comparison to e.g. this <a href="https://www.zeit.de/wissen/gesundheit/2020-03/coronavirus-deutschland-infektionen-faelle-verbreitung-epidemie-karte" target="_blank" rel="noopener">map by ZEIT Online</a>. The exact German locations of coronavirus cases came from <strong>René Engmann and his website <a href="https://www.coronavirus.jetzt/" target="_blank" rel="noopener">coronavirus.jetzt</a>.</strong> He and his team collected the data <a href="https://github.com/iceweasel1/COVID-19-Germany" target="_blank" rel="noopener">in this Github repo.</a> and were amazingly quick at updating the data about Germany until they stopped doing so on March 11th. We still left the out-dated maps in here.</p>

jgehrcke:
I also found that zeit.de is right now the best source in Germany. I think they collaborate with individual state ministries and therefore the numbers are pretty official. They are however much more real-time than the numbers published by Robert Koch Institute (the RKI needs about a day before they process and publish).

jgehrcke:
I am going to help with that @lazd -- my email address is jgehrcke@googlemail.com

jgehrcke:
Unfortunately I could not reach the people behind https://www.coronavirus.jetzt/. Right now zeit.de provides a larger count than coronavirus.jetzt. Given that zeit.de is one of the largest newspapers in Germany (of quality) I will proceed under the assumption that zeit.de is one of the best data sources to bet on for the coming hours, days, and weeks.

jgehrcke:
Here we go:

```
$ curl https://covid19-germany.appspot.com/now 2> /dev/null| jq
{
  "last_update_from_source_iso8601": "2020-03-17T20:03:41+00:00",
  "source": "zeit.de",
  "total_cases_confirmed_until_now": 9293
}
```

[...]

PR #196, opened 2020-03-20
feat: [Germany] add DE-* state scrapers, including timeseries

jgehrcke:
Follow-up from #79.
See commit messages!

I tested this on my machine with

```
$ ± yarn start --date=2020-3-12 --location "DE-BY, DEU"
```
(for checking if time series code path works)

and with
```
± yarn start --location "DE-BY, DEU"
```
to test if the 'current' code path works.

Once we settle on a solution we can easily add more of the same for the individual states in Germany (15 more of them).


lib/fetch.js
```diff
@@ -36,7 +36,7 @@ export const fetch = async (url, type, date = process.env.SCRAPE_DATE || datetim
 
   const cachedBody = await caching.getCachedFile(url, type, date);
 
-  if (cachedBody === caching.CACHE_MISS || alwaysRun) {
+  if (cachedBody === caching.CACHE_MISS || cachedBody === caching.RESOURCE_UNAVAILABLE || alwaysRun) {
```
lazd: @qgolsteyn thoughts?

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,59 @@
+import * as fetch from '../../../lib/fetch.js';
+import * as datetime from '../../../lib/datetime.js';
+
+// Set county to this if you only have state data, but this isn't the entire state
+// const UNASSIGNED = '(unassigned)';
+
+const scraper = {
+  country: 'DEU',
+  state: 'DE-BY', // ISO 3166 notation
+  url: 'https://covid19-germany.appspot.com/',
```
lazd: This should be `scraper.source.url`

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,59 @@
...
+  ],
+  sources: [
+    {
+      name: 'Robert Koch-Institut, Bavarian health ministry',
+      description: 'Fresh data obtained from Bavarian health ministry by ZEIT ONLINE'
+    }
+  ],
+  async scraper() {
+    const queryDayString = datetime.getYYYYMD(new Date(process.env.SCRAPE_DATE));
+    const data = await fetch.csv('https://raw.githubusercontent.com/jgehrcke/covid-19-germany-gae/master/data.csv');
```
lazd: This should be `this.url`

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,59 @@
...
+      }
+    } else {
+      // Rely on last entry in time series to be rather fresh.
+      const [lastRow] = data.slice(-1);
+      queryDayCases = lastRow['DE-BY_cases'];
+      queryDayDeaths = lastRow['DE-BY_deaths'];
+    }
+    return {
+      country: 'DEU',
+      state: 'DE-BY',
```
lazd: `country` and `state` can be omitted here -- they are automatically added if they are defined on the `scraper` object.

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,59 @@
...
+      cases: parseInt(queryDayCases, 10),
+      deaths: parseInt(queryDayDeaths, 10),
+      recovered: undefined, // no credible data from Germany yet
```
lazd: OK, you can drop this completely then -- we don't pass anything at all if the data is not available.

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,59 @@
...
+const scraper = {
+  country: 'DEU',
+  state: 'DE-BY', // ISO 3166 notation
+  url: 'https://covid19-germany.appspot.com/',
+  type: 'json',
+  timeseries: true,
+  curators: [
+    {
+      name: 'Dr. Jan-Philip Gehrcke',
+      email: 'jgehrcke@googlemail.com'
```
lazd: Thank you for curating this data!

jgehrcke:
Thanks for the kind review @lazd -- I pushed a commit implementing the requested changes.

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+  ],
+  sources: [
+    {
+      name: 'Robert Koch-Institut, Bavarian health ministry',
+      description: 'Fresh data obtained from Bavarian health ministry by ZEIT ONLINE'
```
lazd: Does there happen to be a URL for this? It would be good to add it.

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,56 @@
...
+      // Get historical data for a specific day.
+      for (const row of data) {
+        const sampleDayString = datetime.getYYYYMD(new Date(row.time_iso8601));
+        if (sampleDayString === queryDayString) {
+          queryDayCases = row['DE-BY_cases'];
+          queryDayDeaths = row['DE-BY_deaths'];
+        }
+      }
+    } else {
+      // Rely on last entry in time series to be rather fresh.
```
lazd: This makes sense. We've done this with the rest of them. However, if the scrape date is matched, and the requested date is not in the timeseries, we should also use the latest to avoid errors. It will be backfilled when the timeseries is updated, so it's OK.

See ITA and BRA, they both do this.

scrapers/DEU/DE-BY/index.js
jgehrcke: Sure. Thanks. Can I add a `url` property here?

I don't want to take too much credit here, but I think a good URL would be https://github.com/jgehrcke/covid-19-germany-gae -- in the README I tried to explain rather in-depth which sources are used for what.

scrapers/DEU/DE-BY/index.js
jgehrcke: Not sure I understand 

>  if the scrape date is matched

Pseudo code as in

`if SCRAPE_DATE AND SCRAPE_DATE not in dataset THEN return newest datapoint`?

That would probably lead to artifacts for scrape dates too far in the past. mhm, but I probably misunderstood!

lib/fetch.js
qgolsteyn: Hmm, we might be reporting data with misleading dates with this change. Eg, if we are asking for 2020-03-18 and we don't have that in cache, we would fetch data which has been updated today 2020-03-20 and report it as if it was from 2020-03-18.

lazd: Yeah, let's remove this change.

scrapers/DEU/DE-BY/index.js
lazd: Yes, you absolutely can add a URL there, if it details where the data is from, it's relevant!

scrapers/DEU/DE-BY/index.js
lazd: so it goes like this:

```
latestData = dataset(LATEST_DATE)
if LATEST_DATE is before SCRAPE_DATE
  angryErrorMessage
  use(latestData)
else
  if datasetContains(SCRAPE_DATE)
    use dataset(SCRAPE_DATE)
```

This might be the best example, it errors out when data is missing: https://github.com/lazd/coronadatascraper/blob/master/scrapers/CHE/zurich.js#L15-L28

This one doesn't check that the data is there: https://github.com/lazd/coronadatascraper/blob/master/scrapers/ITA/index.js#L14-L24

Neither does this one: https://github.com/lazd/coronadatascraper/blob/master/scrapers/BRA/index.js#L1104-L1114

lib/fetch.js
jgehrcke: Thanks. I am dropping the commit now. Let's keep in mind that without that path I could not test my patch locally with ` yarn start --date=2020-3-12 --location "DE-BY, DEU"`.

Because the fetcher returned `null` w/o even consulting the external source.

I am going to remove the commit, but leave the commit message here for posterity:
```
during testing I ran into

± yarn start --date=2020-3-12 --location "DE-BY, DEU"
yarn run v1.22.4
$ node cli.js --date=2020-3-12 --location 'DE-BY, DEU'
hourglass_flowing_sand Scraping data for 2020-3-12...
  warning Cannot go back in time to get https://covid19-germany.appspot.com/timeseries/DE-BY/cases, no cache present
  warning Cannot go back in time to get https://covid19-germany.appspot.com/timeseries/DE-BY/deaths, no cache present
  x Error processing DE-BY, DEU:  TypeError: Cannot read property 'data' of null
hourglass_flowing_sand Scraping data for 2020-2-26...
  warning Cannot go back in time to get https://covid19-germany.appspot.com/timeseries/DE-BY/cases, no cache present
  warning Cannot go back in time to get https://covid19-germany.appspot.com/timeseries/DE-BY/deaths, no cache present
  x Error processing DE-BY, DEU:  TypeError: Cannot read property 'data' of null

That means that no cache was present
and the data was still not attempted
to be fetched from the remote source.

This patch made my test command work.
```

scrapers/DEU/DE-BY/index.js
jgehrcke: Okay, there are three special cases we'd like to handle consciously:
- two special cases where queryDate is before or after the time/ interval covered by the timeseries.
- special case where time series has a hole.

Patch coming.

jgehrcke:
Okay. I rebased on upstream/master, removed my `cachedBody === caching.CACHE_MISS || cachedBody === caching.RESOURCE_UNAVAILABLE || alwaysRun)` patch from the branch (although I kept it locally for testing), added a source url, and am now explicitly handling these three cases:

- SCRAPE_DATE is before time interval covered by the timeseries: error
- SCRAPE_DATE is after time interval covered by the timeseries: use last sample, angry log msg
- SCRAPE_DATE within time series interval, but sample missing: error

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,79 @@
...
+    {
+      name: 'Robert Koch-Institut, Bavarian health ministry',
+      description: 'Fresh data obtained from Bavarian health ministry by ZEIT ONLINE',
+      url: 'https://github.com/jgehrcke/covid-19-germany-gae'
+    }
+  ],
+  async scraper() {
+    const data = await fetch.csv(this.url);
+
+    function rowToResult(row) {
```
lazd: We've been putting any functions or objects required on the scraper object itself, prefixed with `_`. Just as a matter of consistency, you can move this up to the scraper object:

```
_rowToResult(row) { ...},
```

Then call it with `this._rowToResult()`

jgehrcke: woop, :rocket: fast feedback, commit coming

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,79 @@
...
+      return {
+        cases: parseInt(row['DE-BY_cases'], 10),
+        deaths: parseInt(row['DE-BY_deaths'], 10),
+        coordinates: [11.497, 48.79],
+        population: 13 * 10 ** 6
+      };
+    }
+
+    // location string for log and error msgs.
+    const loc = `${this.country}, ${this.state}`;
```
lazd: You can use `geography.getName(this)` -- it works on any object with the relevant properties (`state`, `country`, etc)

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+    // Handle the two special cases where queryDate is before or after the time
+    // interval covered by the timeseries.
+    if (queryDate > lastDateInTimeseries) {
+      console.error(`  🚨 timeseries for ${loc}: SCRAPE_DATE ${queryDayString} is newer than last sample time ${lastDateInTimeseries}. Use last sample anyway`);
+      return rowToResult(lastRow);
+    }
+
+    if (queryDate < firstDateInTimeseries) {
+      throw new Error(`Timeseries starts later than SCRAPE_DATE ${queryDayString}`);
```
lazd:
```suggestion
      throw new Error(`Timeseries for ${loc} starts later than SCRAPE_DATE ${queryDayString}`);
```

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+    // Search through time series for the query date.
+    for (const row of data) {
+      if (datetime.getYYYYMD(new Date(row.time_iso8601)) === queryDayString) {
+        return rowToResult(row);
+      }
+    }
+
+    // Special case where time series has a hole.
+    throw new Error(`Timeseries does not contain a sample for SCRAPE_DATE ${queryDayString}`);
```
lazd: Fantastic!

scrapers/DEU/DE-BY/index.js
jgehrcke: I had that, but in test I then saw duplication in terminal:
```
Error processing DE-BY, DEU:  Error: Timeseries for DEU, DE-BY starts later than SCRAPE_DATE 2020-3-7
``` 
:)
still want the change?

jgehrcke: adding that back in, better safe than sorry.

lazd: No it's fine as-is, actually! No need to add it

jgehrcke:
Okay, PTAL!

In tests I have seen these two error msgs

```
🚨 timeseries for DE-BY, DEU: SCRAPE_DATE 2020-4-11 is newer than last sample time 2020-3-20. Use last sample anyway
```

```
❌ Error processing DE-BY, DEU:  Error: Timeseries starts later than SCRAPE_DATE 2020-2-28
```

lazd:
This looks excellent, a fine example of how we should handle timeseries data. We need to document and codify the strategy here. 

THANK YOU!

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,77 @@
...
+  _rowToResult: row => {
+    return {
+      cases: parseInt(row['DE-BY_cases'], 10),
+      deaths: parseInt(row['DE-BY_deaths'], 10),
+      coordinates: [11.497, 48.79],
+      population: 13 * 10 ** 6
+    };
+  },
+  async scraper() {
+    const data = await fetch.csv(this.url);
```
lazd: If you pass false here, per-day cache is defeated and it uses the local `./cache/` directory. This is required for timeseries since you don't want to refetch the URL or use cached data ever. Sorry I didn't catch this sooner!
```suggestion
    const data = await fetch.csv(this.url, false);
```

jgehrcke: Thanks! No worries! Pushed a change.

jgehrcke: (man, I just realize: this GitHub code change suggestion feature looks really cool, it's new to me -- next time I will try to just press the button :).

jgehrcke:
> This looks excellent, a fine example of how we should handle timeseries data. We need to document and codify the strategy here.

Thanks for the kind words.

As I said elsewhere, this project's README and website and codebase are excellent, too, especially measured by how quickly all this emerged. Collaboration experience now is a joy, too. A pleasure, thank you for all the great work you put into this, seriously @lazd!

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,77 @@
...
+      cases: parseInt(row['DE-BY_cases'], 10),
+      deaths: parseInt(row['DE-BY_deaths'], 10),
+      coordinates: [11.497, 48.79],
+      population: 13 * 10 ** 6
+    };
+  },
+  async scraper() {
+    const data = await fetch.csv(this.url, false);
+
+    // Rely on dataset to be sorted by time, in direction past -> future.
```
camjc: NaBD but we could explicitly sort it we find it necessary/defensive

scrapers/DEU/DE-BY/index.js
```diff
@@ -0,0 +1,77 @@
...
+      console.error(`  🚨 timeseries for ${geography.getName(this)}: SCRAPE_DATE ${queryDayString} is newer than last sample time ${lastDateInTimeseriesString}. Use last sample anyway`);
+      return this._rowToResult(lastRow);
+    }
+
+    if (queryDate < firstDateInTimeseries) {
+      throw new Error(`Timeseries starts later than SCRAPE_DATE ${queryDayString}`);
+    }
+
+    // Search through time series for the query date.
+    for (const row of data) {
```
camjc: This loop feels semantically like an array.find?

noleti:
I wrote the following re-write of this patch, to be placed in the DEU folder directly. It currently misses location and population data, but that should be easy to hard-code. It now scrapes data for all 16 states from that one API call. The date check thing is taken from ITA parser. Pasting here because I don't know how to fork @jgehrcke's fork of your repo.
[indexGAE.js.txt](https://github.com/lazd/coronadatascraper/files/4365082/indexGAE.js.txt)

noleti:
I updated https://github.com/lazd/coronadatascraper/pull/226 now with the script I attached above. It's based on  @jgehrcke's pull request here, just extended to all states and overall rewritten.

jgehrcke:
> I updated #226 now with the script I attached above. It's based on @jgehrcke's pull request here, just extended to all states and overall rewritten.

Hey Ole. After the last exchange about sharing code across scrapers I noted "will get to that work after getting a bit of sleep. Great work everyone". Well. I actually feel a little bad about that situation that I wake up to. Emotionally, I mean. Because I would kind of obviously love to pick up the work where left off before. Of course it's also great to _actually_ collaborate on code, but an approach like "and overall rewritten" does not make collaboration easy, it makes it technically and emotionally difficult. What's your plan?

noleti:
@jgehrcke sorry, didn't mean to take this away from you. I'm actually not really sure how to collaborate on a pull request, as noted in yours. I didn't want to take over/ this away from you, that's also why I posted the file on your pull request first. I'm still perfectly fine if your pull request gets merged, just got exited about that simpler API endpoint you used, which simplified things a lot.

jgehrcke:
@noleti thank you for your words and this conversation. I think I understand.

Will pick up the work on this again now, will also look at your patches of course. Reporting back later. Have a good Sunday, too! :sun_with_face:

scrapers/DEU/DE-BY/index.js
jgehrcke: For the record: in this case it's a strong guarantee by the dataset used, and I only wanted to make it obvious with this comment that the code relies on that (makes use of the guarantee).

jgehrcke:
We're moving fast in this repo. I rebased on latest master and caught up with us having moved around files/dirs quite a bit. Retained the same commits as before for keeping review work easy, added one towards sharing code across scrapers.

work in progress.

jgehrcke:
One important technical question:

Do we want to have one DEU scraper, and then the 16 individual state scrapers or "just the" 16 individual state scrapers?

I see good arguments for both, need help making an informed decision.

jgehrcke:
@qgolsteyn  could it be that the CI failures of the last three commits are as of your very last commits to master?

jgehrcke:
Okay, rebased again on current upstream/master, force-pushing now. For posterity, CI for the "Add DE-MV scraper" was green above, see this build: https://github.com/lazd/coronadatascraper/runs/525636592

Wen I force-push now, I expect that green checkmark to disappear.

jgehrcke:
> Okay, rebased again on current upstream/master, force-pushing now. For posterity, CI for the "Add DE-MV scraper" was green above, see this build: https://github.com/lazd/coronadatascraper/runs/525636592

Clarified on Slack. This was green because scrapers were not tested. Scrapers are tested again now, and this picked up the _shared.js as a regular scraper.

@qgolsteyn is helping now, to get the _shared.js ignored.

jgehrcke:
> @qgolsteyn is helping now, to get the _shared.js ignored.

@qgolsteyn pushed the relevant commit, rebasing, force-push upcoming

src/events/crawler/tasks/fetchSources.js
```diff
@@ -1,11 +1,16 @@
...
 
 export default async args => {
   console.log(`⏳ Fetching scrapers`);
   const scrapers = join(__dirname, '..', 'scrapers', '**', '*.js').replace(/\\/g, '/');
   let filePaths = await fastGlob([scrapers]);
   filePaths = filePaths.filter(file => !file.endsWith('.test.js'));
+
+  // Enable shared code files in the scrapers directory w/o treating them
+  // as scrapers. See #196.
+  filePaths = filePaths.filter(file => !basename(file).startsWith('_'));
```
jgehrcke: CC @qgolsteyn

src/events/crawler/scrapers/DEU/_shared.js
```diff
@@ -0,0 +1,50 @@
...
+
+  // Search through time series for the query date.
+  for (const row of data) {
+    if (datetime.getYYYYMD(new Date(row.time_iso8601)) === queryDayString) {
+      return this._rowToResult(row);
+    }
+  }
+
+  // Special case where time series has a hole.
+  throw new Error(`Timeseries does not contain a sample for SCRAPE_DATE ${queryDayString}`);
```
jgehrcke: Note to reviewers: I left this basically untouched after the last round of review, we were pretty happy with this core scraping method.

The last changes were all about rebasing to current master (moving around files etc), and then adding 15 other state-level parsers all using the same core scraping method function, the one here.

jgehrcke:
@lazd @qgolsteyn if we have another quick review cycle then maybe we can then still land this today (11 pm here) :) -- btw, which timezone are you in?

src/events/crawler/scrapers/DEU/DE-BB/index.js
```diff
@@ -0,0 +1,44 @@
...
+      name: 'Dr. Jan-Philip Gehrcke',
+      email: 'jgehrcke@googlemail.com',
+      url: 'https://gehrcke.de',
+      github: 'jgehrcke'
+    }
+  ],
+  scraper: defaultScraperDEU,
+  _rowToResult: row => {
+    return {
+      cases: parseInt(row['DE-BB_cases'], 10),
```
qgolsteyn: Use `parse.number()` from `lib/parse.js`

src/events/crawler/scrapers/DEU/DE-BB/index.js
```diff
@@ -0,0 +1,44 @@
...
+      deaths: parseInt(row['DE-BB_deaths'], 10),
```
qgolsteyn: Use `parse.number()` from `lib/parse.js`. Similar issues in all scrapers.

src/events/crawler/scrapers/DEU/DE-BE/index.js
```diff
@@ -0,0 +1,43 @@
...
+      name: 'Dr. Jan-Philip Gehrcke',
+      email: 'jgehrcke@googlemail.com',
+      url: 'https://gehrcke.de',
+      github: 'jgehrcke'
+    }
+  ],
+  scraper: defaultScraperDEU,
+  _rowToResult: row => {
+    return {
+      cases: parseInt(row['DE-BE_cases'], 10),
```
qgolsteyn: Use `parse.number()` from `lib/parse.js`

src/events/crawler/scrapers/DEU/DE-BE/index.js
```diff
@@ -0,0 +1,43 @@
...
+      deaths: parseInt(row['DE-BE_deaths'], 10),
```
qgolsteyn: Use `parse.number()` from `lib/parse.js`

src/events/crawler/tasks/fetchSources.js
qgolsteyn: Looks good!

jgehrcke:
> Looks good! Ton of effort here! Would it be possible to move all duplicate fields (country, url, timeseries, etc.) to _shared.js and spread them into each file? This will keep all the scraper DRY.

I feel you! I was there, trying to do exactly that. Defining and exporting more objects in `_shared.js` than the scrape function led to complications in my attempt, said import export hell. These complications are probably easy to get around, if you know what you're doing. I am not a JS pro, and the gazillion ways of exporting and importing still confuse me like hell.  I couldn't quite manage to get done what I wanted, in one of my attempt I broke the `this.` association, and that frustrated me :). Will try to give this another attempt tomorrow. There must be a clean way :).

qgolsteyn:
@jgehrcke I helped you move the schema to _shared. Once you address the other comments, I can merge this!

jgehrcke:
@qgolsteyn lovely, thanks for helping. Looks very elegant! Seeing this spread in action makes it pretty obvious that this is the solution to go. :)

jgehrcke:
Thinking out loud: Should I now continue to merge master into this branch here like you did, or should I try rebasing again?

jgehrcke:
Let's have another look, but this should be good to go. :rocket: 
I know there are many eyes on this, glad when Germany is finally represented in this project. Thanks for all the help.

qgolsteyn:
No need to merge in master unless it conflicts. Let me have a look at this.

jgehrcke:
@qgolsteyn thanks. I want to again say _thank you_ for the excellent work, and for the splendid contribution experience.

PR #198, opened 2020-03-20
feat: Add GA County Case Data

jeffchenoweth:


scrapers/USA/GA/index.js
```diff
@@ -0,0 +1,29 @@
...
+    const $table = $('table:contains(County):contains(Cases)');
+    const $trs = $table.find('tbody > tr');
+    $trs.each((index, tr) => {
+      const $tr = $(tr);
+      const county = transform.addCounty(parse.string($tr.find('td:first-child').text()));
+      const cases = parse.number($tr.find('td:last-child').text());
+      counties.push({ county, cases });
+    });
+
+    return counties;
```
lazd: Add the rollup for GA
```suggestion
    counties.push(transform.sumData(counties));
    return counties;
```

scrapers/USA/GA/index.js
```diff
@@ -0,0 +1,29 @@
...
+    const $ = await fetch.page(this.url);
+    const counties = [];
+    const $table = $('table:contains(County):contains(Cases)');
+    const $trs = $table.find('tbody > tr');
+    $trs.each((index, tr) => {
+      const $tr = $(tr);
+      const county = transform.addCounty(parse.string($tr.find('td:first-child').text()));
+      const cases = parse.number($tr.find('td:last-child').text());
+      counties.push({ county, cases });
+    });
```
lazd: Does this page contain every single county in GA? Otherwise, you can follow the patterns in OH (pull county list from Wikipedia, use `transform.addEmptyRegions()` to add those with zero cases).

lazd: I went ahead and added it!

lazd:
So @jeffchenoweth totally forgot to mention that we added `source` and `maintainer` fields -- they're not required, but if you put yourself as `maintainer` and agree to maintain the scraper, you'll get credit on the site. If that sounds good, do it like this: https://github.com/lazd/coronadatascraper/blob/master/scrapers/BRA/index.js#L15-L24

```
source:  {
  name: 'Georgia Department of Public Health'
}
```

If you want to send along another PR with that info, I can get it merged in and deployed tonight!

PR #200, opened 2020-03-20
county level scraper for MO

paulboal:
This is a county-level scraper for MO from the public health website that was already documented: https://health.mo.gov/living/healthcondiseases/communicable/novel-coronavirus/


scrapers/USA/MO/index.js
```diff
@@ -0,0 +1,178 @@
...
+
+    const $trs = $table.find('tr:not(:first-child)');
+    $trs.each((index, tr) => {
+      const $tr = $(tr);
+      let countyName = parse.string($tr.find('td:nth-child(1)').text());
+      countyName = this._countyMap[countyName] || countyName;
+
+      const casesState = parse.number($tr.find('td:nth-child(2)').text()) || 0;
+      const casesOther = parse.number($tr.find('td:nth-child(3)').text()) || 0;
+      if (countyName !== 'TBD') {
```
lazd: For `TBD`, can you add it to another object `{ county: UNASSIGNED, cases: cases)`?

scrapers/USA/MO/index.js
```diff
@@ -0,0 +1,178 @@
...
+          counties[countyName] += casesState + casesOther;
+        } else {
+          counties[countyName] = casesState + casesOther;
+        }
+      }
+    });
+
+    const countiesList = [];
+
+    for (const c in counties) {
```
lazd: `transform.objectToArray()`

scrapers/USA/MO/index.js
paulboal: Got it.  Done.

scrapers/USA/MO/index.js
paulboal: Didn't see that.  Great! and Done.

scrapers/USA/MO/index.js
```diff
@@ -0,0 +1,171 @@
...
+    $trs.each((index, tr) => {
+      const $tr = $(tr);
+      let countyName = parse.string($tr.find('td:nth-child(1)').text());
+      countyName = this._countyMap[countyName] || countyName;
+
+      const casesState = parse.number($tr.find('td:nth-child(2)').text()) || 0;
+      const casesOther = parse.number($tr.find('td:nth-child(3)').text()) || 0;
+      if (countyName !== 'UNASSIGNED') {
+        countyName = transform.addCounty(countyName);
+      }
```
lazd: So for this, you previously had stuff matching on `TBD` -- we don't want to ignore that, it needs to be added into a separate object with `{ county: UNASSIGNED }` -- the constant from above (see the code comment at the top of the file).

paulboal: Thanks for the clarification.  Now I get it.

lazd:
In my haste of merging your excellent contribution, I forgot to ask that you add yourself as the maintainer if you'd like to be credited on the site and are willing to maintain it (like done here): https://github.com/lazd/coronadatascraper/blob/master/scrapers/BRA/index.js#L15-L24

```
source:  {
  name: 'Missouri Department of Health and Senior Services'
}
```

PR #203, opened 2020-03-20
Favicons and logo tweak

thatneat:
I noticed there was no favicon, and then pulled on the thread. Here's what I ended up changing:

- converted the inline SVG logo to a file for easier reuse. This required removing the 30-degree rotation on the background gradient, which is a change I couldn't actually detect when I was looking for it. This may be because it's unsupported? Either way, `gradientTransform="rotate(30)" was preventing the SVG file from being opened, and when I removed it, everything looked the same.

Based on this SVG (or actually, a .png export of it), I used realfavicongenerator.net to generate a full complement of favicon-type assets, including:
- favicon (in root directory for maximum compatibility)
- android/iOS/Windows shortcut icons
- Safari pinned tab icon

I tried to make reasonable choices for transparency, background colors, drop shadow, etc. in the favicon generator UI.

favicon and refactored logo in Firefox:
image


lazd:
@NateBaldwinDesign if this looks good to you, let's merge!

NateBaldwinDesign:
Looks good to me

PR #206, opened 2020-03-20
county level scraper for OK

paulboal:
This is a county-level scraper for OK from the public health website that was already documented: https://www.ok.gov/health/Prevention_and_Preparedness/Acute_Disease_Service/Disease_Information/Coronavirus_Disease_2019/Oklahoma_Response_to_Coronavirus_Disease_2019/index.html


scrapers/USA/OK/index.js
```diff
@@ -0,0 +1,137 @@
...
+          counties[countyName] += casesState;
+        } else {
+          counties[countyName] = casesState;
+        }
+      }
+    });
+
+    const countiesList = [];
+
+    for (const c in counties) {
```
lazd: We have `transform.objectToArray()` that does this for ya

paulboal: Got it.

lazd:
I made 2 tiny changes and renamed the functions I just refactored, and this is going in! Thank you!!

lazd:
In my haste of merging your excellent contribution, I forgot to ask that you add yourself as the maintainer if you'd like to be credited on the site and are willing to maintain it (like done here): https://github.com/lazd/coronadatascraper/blob/master/scrapers/BRA/index.js#L15-L24

```
source:  {
  name: 'Oklahoma State Department of Health'
}
```

PR #209, opened 2020-03-21
feat: split scrapeData into sub tasks

qgolsteyn:
This PR splits scrapeData into subtasks. This should help us add additional steps to scrapeData and test them effectively.


tasks/scrapeData/cleanLocations.js
```diff
@@ -0,0 +1,39 @@
+const scraperVars = ['type', 'timeseries', 'headless', 'ssl', 'priority', 'aggregate'];
+
+/*
+  Remove "private" object properties
+*/
+const removePrivate = location => {
```
camjc: Same as in the next file.

Also may as well do `if empty or underscore` rather than write the `delete` twice

tasks/scrapeData/dedupData.js
```diff
@@ -0,0 +1,113 @@
...
+    const otherLocation = seenLocations[locationName];
+
+    if (otherLocation) {
+      // Take rating into account to break ties
+      const thisPriority = getPriority(location) + location.rating / 2;
+      const otherPriority = getPriority(otherLocation) + otherLocation.rating / 2;
+
+      if (otherPriority === thisPriority) {
+        console.log('⚠️  %s: Equal priority sources choosing %s (%d) over %s (%d) arbitrarily', locationName, location.url, thisPriority, otherLocation.url, otherPriority);
+        // Kill the other location
```
camjc: Potentially delete is a nicer word since we're talking about viruses 🤷‍♂

lazd: Oh snap, good call. That was my wording.

tasks/scrapeData/cleanLocations.js
lazd: Good call, though refactoring I ended up with separate conditionals, they should be combined to one expression.

tasks/scrapeData/dedupData.js
```diff
@@ -0,0 +1,113 @@
...
+      failed = true;
+    }
+  }
+  return failed ? crosscheckReport : false;
+}
+
+/*
+  Remove "private" object properties
+*/
+function removePrivate(data) {
```
lazd: @qgolsteyn can we share these? They could live in `lib/transform.js`

tasks/scrapeData/dedupData.js
```diff
@@ -0,0 +1,113 @@
...
+      break;
+    }
+  }
+  return exists;
+}
+
+/*
+  Get the priority of a location
+*/
+const getPriority = location => {
```
lazd: @qgolsteyn I believe this already lives in `lib/geography.js`

tasks/scrapeData/dedupData.js
```diff
@@ -0,0 +1,113 @@
...
+  return location.priority !== undefined ? location.priority : 0;
+};
+
+const dedupLocations = args => {
```
lazd: Should be `dedupeLocations`

tasks/scrapeData/dedupData.js
```diff
@@ -0,0 +1,113 @@
...
+        const stippedOtherLocation = removePrivate(otherLocation);
+        if (!existsInCrosscheckReports(stippedOtherLocation, crosscheckReports[locationName])) {
+          crosscheckReports[locationName].push(stippedOtherLocation);
+        }
+      }
+    }
+    seenLocations[locationName] = location;
+  }
+
+  return { ...args, deDuped, crosscheckReports };
```
lazd: Yay for spread

tasks/scrapeData/normalizeLocations.js
```diff
@@ -0,0 +1,83 @@
...
+  Normalize the state as a 2-letter string
+*/
+const toUSStateAbbreviation = str => {
+  return usStates[str] || str;
+};
+
+/*
+  Normalize the state as a 2-letter string
+*/
+const toISO3166Alpha3 = string => {
```
lazd: This and other functions were moved to `lib/geography.js`

tasks/scrapeData/normalizeLocations.js
```diff
@@ -0,0 +1,83 @@
...
+  'Congo (Brazzaville)': 'COG',
+  Tanzania: 'TZA',
+  'The Bahamas': 'BHS',
+  'Gambia, The': 'GMB'
+};
+
+/**
+  Normalize the state as a 2-letter string
+*/
+const toUSStateAbbreviation = str => {
```
lazd: This and other functions were moved to `lib/geography.js`

tasks/scrapeData/normalizeLocations.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+  console.warn('⚠️  Could not find country code for', string);
+  return string;
+};
+
+/*
+  Calculates active cases from location data
+ */
+const getActiveFromLocation = location => {
```
lazd: This and other functions were moved to `lib/geography.js`

tasks/scrapeData/runScraper.js
```diff
@@ -0,0 +1,131 @@
+import * as datetime from '../../lib/datetime.js';
+import * as geography from '../../lib/geography.js';
+
+const numericalValues = ['cases', 'tested', 'recovered', 'deaths', 'active'];
```
lazd: I was thinking we should put this in a file somewhere, but it was late and I couldn't think of where it should live...

It's used in a few random places across the project (CSV generation, validation, etc).

Thoughts on a name? `lib/schema.js` sounds wrong

qgolsteyn: I think it definitely needs to live somewhere, and `lib/schema.js` sounds like a good name

tasks/scrapeData/dedupData.js
qgolsteyn: Would prefer keeping them to this file as they are not used anywhere else for now. Seems more of a private method.

cli.js
```diff
@@ -1,6 +1,8 @@
 const imports = require('esm')(module);
 
-const generate = imports('./index.js').default;
+const generate = imports('./tasks/index.js').default;
 const argv = imports('./lib/cliArgs.js').default;
 
-generate(argv.date, argv);
+generate(argv.date, argv).catch(e => {
```
camjc: Does this get swallowed otherwise?

qgolsteyn: Yup

lib/transform.js
```diff
@@ -1,6 +1,32 @@
...
+*/
+export const removePrivate = data => {
+  for (const [prop, value] of Object.entries(data)) {
+    if (value === '') {
+      delete data[prop];
+    }
+    // Remove "private" fields
+    if (prop[0] === '_') {
+      delete data[prop];
+    }
```
camjc:
```suggestion
    if (value === '' || prop[0] === '_') {
      delete data[prop];
    }
```

ryanblock:
Stoked you're on this!

tasks/scrapeData/dedupeLocations.js
```diff
@@ -0,0 +1,106 @@
...
+      failed = true;
+    }
+  }
+  return failed ? crosscheckReport : false;
+}
+
+/*
+  Remove "private" object properties
+*/
+function removePrivate(data) {
```
camjc: Should we just import the one from transform.js or how best to DRY?

tasks/scrapeData/rateLocations.js
```diff
@@ -91,4 +94,40 @@ function calculateRating(info) {
...
+  const sourcesByURL = {};
+  for (const location of locations) {
+    const sourceObj = { ...location._scraperDefinition };
+    for (const prop of sourceProps) {
+      if (location[prop] !== undefined) {
+        sourceObj[prop] = location[prop];
+      }
+    }
+    for (const prop in sourceObj) {
+      if (prop[0] === '_') {
```
camjc: also repetitious with transform.js remove private thing

tasks/scrapeData/runScraper.js
camjc: Yep, schema or constants or something, good to move to a file 👍

tasks/scrapeData/runScraper.js
```diff
@@ -0,0 +1,131 @@
...
+    // Important: this prevents SSL from failing
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+  }
+  if (typeof location.scraper === 'function') {
+    return location.scraper();
+  }
+  if (rejectUnauthorized) {
+    delete process.env.NODE_TLS_REJECT_UNAUTHORIZED;
+  }
+  if (typeof location.scraper === 'object') {
```
camjc: lots of JS stuff is an object, could be better to use something more robust eventually (lodash isPlainObject for example)

tasks/scrapeData/runScraper.js
```diff
@@ -0,0 +1,131 @@
...
+    if (options.location) {
+      if (geography.getName(location) !== options.location) {
+        continue;
+      }
+    }
+    if (options.skip) {
+      if (geography.getName(location) === options.skip) {
+        continue;
+      }
+    }
```
camjc:
```suggestion
    if (options.skip && geography.getName(location) === options.skip) {
      continue;
    }
```

tasks/scrapeData/runScraper.js
```diff
@@ -0,0 +1,131 @@
...
+  const { sources, options } = args;
+
+  const locations = [];
+  const errors = [];
+  for (const location of sources) {
+    if (options.location) {
+      if (geography.getName(location) !== options.location) {
+        continue;
+      }
+    }
```
camjc:
```suggestion
    if (options.location && geography.getName(location) !== options.location) {
      continue;
    }
```

tasks/scrapeData/rateLocations.js
qgolsteyn: I'm going to keep this to avoid changing too much at once. We can remove this in a future PR (which will happen soon I promise!)

camjc:
This is looking great @qgolsteyn! 🌟

tasks/scrapeData/rateLocations.js
qgolsteyn: #212 to track

tasks/scrapeData/runScraper.js
qgolsteyn: Will do when adding unit tests for this file

qgolsteyn:
Alright, I think I covered all the comments so far

PR #210, opened 2020-03-21
Prettier

hyperknot:
- added pretty command to recursively fix all JS files
- set line_lenth to 120
- yarn.lock was left out-of-sync in master, this fixes it

If you are OK with these changes, we should run pretty to reformat the codebase to line_length=120.


hyperknot:
If you are OK with these changes, I'm happy to push the `yarn pretty` run as well, to fix the test, etc. But that'd kill the "files changed" tab in GitHub, so I submitted this version first.

lazd:
So simply changing the line length makes prettier format objects differently? Why can't we keep a huge line length and tell it to format objects in a sane way...

hyperknot:
It cannot, it bases object formatting on line length. So `{a:1, b:2, c:3}` stays on one line but the big objects go multi line.

Actually life with 120 is not that bad, I got used to it when coming from 200.

hyperknot:
I went through the diff for `pretty` with 120 and I believe the codebase definitely gets more readable.

camjc:
FWIW I do 120 at work, seems nice to me 🤷‍♂

PR #213, opened 2020-03-21
Treat missing/NaN values as 0 in WA, USA scraper

chunder:


scrapers/USA/WA/index.js
```diff
@@ -93,8 +93,16 @@ const scraper = {
       const $trs = $table.find('tbody > tr');
       $trs.each((index, tr) => {
         const $tr = $(tr);
-        const cases = parse.number($tr.find('td:nth-child(2)').text());
-        const deaths = parse.number($tr.find('td:last-child').text());
+        let cases = parse.number($tr.find('td:nth-child(2)').text());
+        let deaths = parse.number($tr.find('td:last-child').text());
```
lazd: So with this code, if the text changes to `something`, this will return zero. It's possible that, if the text is truly a non-empty and non-number thing, we want an error.

In that case, this should still work for empty cells, but should still explode if the cell contains non number data:

```suggestion
        let deaths = parse.number(parse.string($tr.find('td:last-child').text()) || 0);
```
Thoughts? Silent failures are scary.

scrapers/USA/WA/index.js
```diff
@@ -93,8 +93,16 @@ const scraper = {
       const $trs = $table.find('tbody > tr');
       $trs.each((index, tr) => {
         const $tr = $(tr);
-        const cases = parse.number($tr.find('td:nth-child(2)').text());
-        const deaths = parse.number($tr.find('td:last-child').text());
+        let cases = parse.number($tr.find('td:nth-child(2)').text());
```
lazd: Go ahead and make the same change here, then remove the `isNan` checks

scrapers/USA/WA/index.js
```diff
@@ -93,8 +93,9 @@ const scraper = {
       const $trs = $table.find('tbody > tr');
       $trs.each((index, tr) => {
         const $tr = $(tr);
-        const cases = parse.number($tr.find('td:nth-child(2)').text());
-        const deaths = parse.number($tr.find('td:last-child').text());
+        const cases = parse.number($tr.find('td:nth-child(2)').text() || 0);
```
lazd:
```suggestion
        const cases = parse.number(parse.string($tr.find('td:nth-child(2)').text()) || 0);
```

PR #216, opened 2020-03-21
MO move county data and add maintainer info

paulboal:
MO moved their county-level data to a new page.  I've fixed that add added my maintainer info to MO and OK.


paulboal:
And now I see how you imported from the maintainers file into the WY scraper file.  I'll clean that up for MO and OK on my next PR.

PR #217, opened 2020-03-21
Fixed Shasta County, CA

jbencina:
Had to convert to headless since standard request wasn't pulling in the number shown in the browser.


lazd:
Ah, I was able to fix this with a new URL. I'll go ahead and close this.

PR #219, opened 2020-03-21
Updated MO Counties

jbencina:
Updated MO. Table no longer contains "County" in the name except for one. Added check to append county to minimize work needed. Also updated URL.

### DIFF
--- a/scrapers/USA/MO/index.js
+++ b/scrapers/USA/MO/index.js
@@ -149,26 +149,27 @@ const scraper = {
     'Worth County',
     'Wright County'
   ],
-  async scraper() {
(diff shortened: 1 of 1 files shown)


lazd:
Hey @jbencina , lemme know if you think we need this one merged. Having trouble reconciling the specific differences between the scrapers, and it seems to work right now as-is. Does this fix a timeseries issue or something?

Thank you!

jbencina:
I think we can drop this? It looks like @paulboal and I both wrote a PR https://github.com/lazd/coronadatascraper/pull/216. It seems we have data in there for 3/20 so it appears to be working as is

PR #220, opened 2020-03-21
Fix maintainer info

paulboal:
Updating maintainer info and fixing MO location -- they moved that data already!


paulboal:
Just saw that @jbencina made a better fix to keep the before 3/20 and after 3/20 locations.  I'm going to close this PR and wait for that one to get merged.

PR #222, opened 2020-03-21
[minor] add AUS VIC

camjc:
Also annoyingly a paragraph from another page :|

### DIFF
--- a/scrapers/AUS/VIC/index.js
+++ b/scrapers/AUS/VIC/index.js
@@ -15,7 +15,7 @@ const scraper = {
     const paragraph = $currentArticlePage('.page-content p:first-of-type').text();
     const { casesString } = paragraph.match(/cases in Victoria to (?<casesString>\d+)./).groups;
     const cases = parse.number(casesString);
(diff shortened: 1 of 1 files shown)


scrapers/AUS/VIC/index.js
```diff
@@ -0,0 +1,22 @@
...
+  priority: 1,
+  async scraper() {
+    const $ = await fetch.page(this.url);
+    const anchors = $('.content ul li a');
+    const currentArticleUrl = anchors[0].attribs.href;
+    const $currentArticlePage = await fetch.page(`https://www.dhhs.vic.gov.au${currentArticleUrl}`);
+    const paragraph = $currentArticlePage('.page-content p:first-of-type').text();
+    const { casesString } = paragraph.match(/cases in Victoria to (?<casesString>\d+)./).groups;
+    const cases = parse.number(casesString);
+    return [{ state: scraper.state, cases }];
```
lazd: Note you can return an object here and it'll work fine.
```suggestion
    return { state: scraper.state, cases };
```

PR #224, opened 2020-03-21
AWS production prep

ryanblock:


rollup.config.js
```diff
@@ -24,10 +24,13 @@ export default {
         contentBase: 'dist',
         port: 3000
       }),
-    !prod && livereload(),
+    !prod && livereload({ watch: 'site' }),
```
lazd: Well that explains a lot. Thank you!

lazd:
Looking at this, and I don't think it's terrible for discovery, `scrapers` → `src/events/crawler/scrapers` isn't the worst rename.

Now I'm wondering if `site/` → `src/site/` should be part of this PR, or do it later? `site/index.js` has to reach out to `'../src/events/crawler/lib/geography.js'` and it might feel more natural as `'../events/crawler/lib/geography.js'`.

We talked about `lib/` → `src/shared/lib`, and that resonates with me.

What about `scripts/` → `src/scripts/`? 

What about `tools/` → `src/tools/`?

@qgolsteyn @camjc @hyperknot, thoughts?

ryanblock:
> Now I'm wondering if site/ → src/site/ should be part of this PR, or do it later? site/index.js has to reach out to '../src/events/crawler/lib/geography.js' and it might feel more natural as '../events/crawler/lib/geography.js'.

I can def do that as a quick followup!

> We talked about lib/ → src/shared/lib, and that resonates with me.

I'll take a look at that – there were some future-facing considerations that paused me from doing it now but I think it's feasible.

> What about scripts/ → src/scripts/?
> What about tools/ → src/tools/?

Nothing to stop us from doing that, can do it at the same time as `src/site` if you like.

lazd:
@ryanblock I say moving everything to `src/` now is likely the best way to keep things consistent. @qgolsteyn thoughts?

qgolsteyn:
I would say from a discoverability standpoint, keeping `scrapers` at the project root would be best. If that's not possible, having it in `src` would still be okay. For `scripts`, this folder deals with CI and local development scripts, probably best to keep it at the project root. For `tools`, most likely a folder to move into `src`

lazd:
@qgolsteyn unfortunately, `scrapers` does have to go deep. We can keep `scripts` at the top level thought.

lazd:
@ryanblock I say we move:

* `tools/` → `src/tools/`
* `site/` → `src/site/`
* `lib/` → `src/shared/lib`

And then merge this bad boy.

hyperknot:
Do we really need to keep scrapers in `/src/events/crawler/scrapers`? This project is 99% about scrapers. Why move it there?

I personally found the original system to be more developer friendly.

ryanblock:
Because the crawler dir will be what executes in the Lambda.

PR #226, opened 2020-03-21
German state scrapers with metadata

noleti:
German State-level scrapers of the https://github.com/jgehrcke/covid-19-germany-gae api. Code is quite redundant, but this allows for state-level optimizations later. supports timeseries.


jgehrcke:
@noleti Hey man thank you big times for doing this, based on https://github.com/jgehrcke/covid-19-germany-gae!! It's so cool to see all this lovely collaboration.

I don't know if you had seen https://github.com/lazd/coronadatascraper/pull/196 -- here I was for starters proposing _one_ scraper out of the 16 that we need, to get feedback on the implementation from @lazd and @qgolsteyn 

I also [commented](https://github.com/lazd/coronadatascraper/pull/196#issue-391569878)

> Once we settle on a solution we can easily add more of the same for the individual states in Germany (15 more of them).

We're still iterating on  https://github.com/lazd/coronadatascraper/pull/196 but that should settle within the next hours, and then I think we should add the other scrapers for the individual German states with _shared_ code. Ha... initiatives everywhere of course create the overhead of potentially duplicating work :).

jgehrcke:
On the other hand, of course, if everyone is cool with the state of _this_ PR here then I am not married to https://github.com/lazd/coronadatascraper/pull/196 (although I wish I had seen this on here earlier, woopsie...).

noleti:
I see you got quite some feedback on #196 already, sorry I missed that. I'm just looking forward to see that data in this API. Honestly, the biggest pain for me was the LON LAT and population lookup for each state. I mainly re-used code from the ITA parser.

noleti:
I noted some issues with my code, so don't merge now.

lazd:
@noleti @jgehrcke please work together to consolidate #229, this PR, and #196 into a single PR. Thank you for putting in the work!

jgehrcke:
> @noleti @jgehrcke please work together to consolidate #229, this PR, and #196 into a single PR. Thank you for putting in the work!

Getting #196 already merged might make this easier but I think @lazd the core question here is: do we have an example for code shared across scrapers? Should we try to do that, or do we want to have copy pasta with quite a bit of duplicated code across the individual german state country scrapers?

#229 uses RKI data which we for now decided against, because for now we'd like to use the more recent data published by individual Gesundheitsaemter in Germany -- exactly what https://github.com/jgehrcke/covid-19-germany-gae provides.

lazd:
@jgehrcke go ahead and create a file inside of `scrapers/DEU/_shared.js`. We'll have to modify `src/events/crawler/tasks/scrapeData/runScraper.js` to not try to execute it, then we should be good to share code.

@qgolsteyn @samirrayani thoughts?

jgehrcke:
> @jgehrcke go ahead and create a file inside of scrapers/DEU/_shared.js. We'll have to modify src/events/crawler/tasks/scrapeData/runScraper.js to not try to execute it, then we should be good to import that file in both scrapers and share code. 

Sounds good, thanks for feedback. Will get to that work after getting a bit of sleep. Great work everyone.

qgolsteyn:
@lazd Yup, ignoring `_*.js` files can be done with no impact on the other scrapers. I think this is a great idea.

noleti:
It's not really clear to me why the test fails, please advise

camjc:
@noleti is it to do with the directory move to src/events/crawler/scrapers ?

noleti:
possibly. as https://github.com/lazd/coronadatascraper/pull/196 was merged now (which uses 16 scrapers, one per state), there is not point in merging this PR (which used one parser for 16 states, using the same API). Please feel free to close the PR in case you agree.

PR #228, opened 2020-03-21
feat: Spain scraper

HerbCaudill:
- Changes the Spain scraper to use a different datasource that includes a time series 
- Adds geojson shapes for Spain's true first-level admin divisions (comunidades autónomas)
- Adds latest official Spain population data

### New dependency: `js-joda`

I've added [js-joda](https://github.com/js-joda/js-joda) as a dependency. I've noticed in the commit log that there have been some struggles with the Javascript `Date` object and time zones; The js-joda `DateLocal` object makes all that go away, as it is a clean date with no time and no timezone. (As time permits I'd like to refactor some of the existing code to use this; for now it's only used in the new Spain scraper.)


lazd:
@hypercube @qgolsteyn thoughts on the `js-joda` bit? It's jumping ahead a bit, and we'll have to refactor it when we do #42, but I'm not oppoosed.

scrapers/ESP/index.js
```diff
@@ -1,38 +1,128 @@
...
+      // but is (maybe?) planning to switch to yyyymmdd.
+      // See https://github.com/datadista/datasets/issues/14
+      // When/if this happens, replace contents of this function with:
+      // return LocalDate.parse(s)
+
+      const [d, m, y] = s.split('/');
+      return LocalDate.parse(`${y}-${m}-${d}`);
+    };
+
+    const mostRecent = dates => dates.sort().pop();
```
lazd: Can you take a look at the way we're doing timeseries checks in https://github.com/lazd/coronadatascraper/pull/196 and make sure you handle the same errors?

* Requested date is before timeseries begin (error)
* Requested date is after latest date in timeseries (use latest date)

lazd:
@HerbCaudill, this is fantastic work. We need to sort out the pattern for handling errors when dates are requested that are outside of the timeseries, and let the team decided on `js-joda`, but this is excellent. Thank you!

scrapers/ESP/index.js
```diff
@@ -1,38 +1,128 @@
...
-// const UNASSIGNED = '(unassigned)';
 
 const scraper = {
   country: 'ESP',
-  url: 'https://opendata.arcgis.com/datasets/48fac2d7de0f43f9af938852e3748845_0.csv',
+  url: 'https://github.com/datadista/datasets/tree/master/COVID%2019',
   priority: 1,
+  headless: false,
+  timeseries: true,
   aggregate: 'state',
```
lazd: Can you cite whoever is curating this with the `curators` field? I understand it's community sources, is there a web page for the community? Do they cite their sources?

scrapers/ESP/index.js
```diff
@@ -1,38 +1,128 @@
...
-
-// Set county to this if you only have state data, but this isn't the entire state
-// const UNASSIGNED = '(unassigned)';
 
 const scraper = {
   country: 'ESP',
-  url: 'https://opendata.arcgis.com/datasets/48fac2d7de0f43f9af938852e3748845_0.csv',
+  url: 'https://github.com/datadista/datasets/tree/master/COVID%2019',
   priority: 1,
+  headless: false,
```
lazd:
```suggestion
```

scrapers/ESP/index.js
```diff
@@ -1,38 +1,128 @@
...
+        const recoveredRow = rawData.recovered.find(sameLocation);
+        return Object.keys(casesRow)
+          .filter(isDate)
+          .map(date => {
+            return {
+              state: location,
+              date: parseDate(date).format(ISO),
+              cases: +casesRow[date] || 0,
+              deaths: +deathsRow[date] || 0,
+              recovered: +recoveredRow[date] || 0
```
lazd: Please use `parse.number()`

scrapers/ESP/index.js
```diff
@@ -1,38 +1,128 @@
...
+      .flatMap(casesRow => {
+        const location = casesRow.CCAA;
+        const sameLocation = d => d.CCAA === location;
+        const deathsRow = rawData.deaths.find(sameLocation);
+        const recoveredRow = rawData.recovered.find(sameLocation);
+        return Object.keys(casesRow)
+          .filter(isDate)
+          .map(date => {
+            return {
+              state: location,
```
lazd: Please use `parse.string()`

scrapers/ESP/index.js
```diff
@@ -1,38 +1,128 @@
...
+    const dates = Object.keys(sampleRow)
+      .filter(isDate)
+      .map(parseDate);
+    const latestDate = mostRecent(dates);
+
+    // use the scrape date, or the latest available
+    const date = (scrapeDate || latestDate).format(ISO);
+
+    // return data from that date
+    const locations = data.filter(d => d.date === date);
```
lazd: In and around here is where you need the logic to handle the errors for requested date out of range.

qgolsteyn:
@lazd Looking at `js-joda`, I would give my support. It fits our use case quite nicely, as `LocalDate` deals with dates only and removes timezone issues, which would be beneficial for our project.

scrapers/ESP/index.js
HerbCaudill: I have some ideas about how we can centralize this logic but don't want to step on anyone's toes if someone's already on it - let me know.

scrapers/ESP/index.js
HerbCaudill: I've added the `curators` field. The sources are Spain's Ministry of Health and Department of National Security, which publishes in PDF form. Is there a good place to note that in the scraper's metadata?

HerbCaudill:
Should be good to go, other than the CI failure - I haven't worked with git submodules before, so don't know how to make it happy. Also don't know if there's a way to tie this PR with https://github.com/lazd/coronavirus-data-sources/pull/8 , which it depends on.

hyperknot:
@HerbCaudill why is joda better then Luxon with a utc timezone? We'll soon need to start handling scraping across timezones, as in we need to know what is the local time of the website we are scraping and were thinking about using Luxon throughout the project.

In my understanding this scraper is "lucky" in that it is already reading data from an other GitHub project. Normally this is not the case when we fetch a government website for example.

In my experience Luxon with 'utc' timezone is working exactly like a local time, I believe it functions like Joda.

camjc:
Never heard of joda but looks pretty good.

lib/parse.js
```diff
@@ -5,7 +5,7 @@ export const number = function(string) {
   if (typeof string === 'number') {
     return string;
   }
-  if (string === '') {
+  if (string === '' || string === undefined) {
```
camjc: Is it possible this will over-coerce? That is, hide some errors we’d want to know about?

package.json
```diff
@@ -81,5 +82,10 @@
       "eslint --fix",
       "git add"
     ]
+  },
+  "prettier": {
+    "printWidth": 500,
```
camjc: Thought e we were going for 120?

scrapers/ESP/index.js
```diff
@@ -1,38 +1,135 @@
...
-import * as rules from '../../lib/rules.js';
-
-// Set county to this if you only have state data, but this isn't the entire state
-// const UNASSIGNED = '(unassigned)';
+import * as parse from '../../lib/parse.js';
 
 const scraper = {
   country: 'ESP',
-  url: 'https://opendata.arcgis.com/datasets/48fac2d7de0f43f9af938852e3748845_0.csv',
+  url: 'https://github.com/datadista/datasets/tree/master/COVID%2019',
```
camjc: Is this a primary source?

scrapers/ESP/index.js
```diff
@@ -1,38 +1,135 @@
...
+    //      // same format
+    //    ]
+    // }
+    // ```
+
+    const data = rawData.cases
+      .filter(casesRow => casesRow.CCAA !== 'Total')
+      .flatMap(casesRow => {
+        const location = casesRow.CCAA;
+        const sameLocation = d => d.CCAA === location;
```
camjc: Nabd Could name the start as a verb

package.json
HerbCaudill: I was just replicating `.editorconfig` settings:
https://github.com/HerbCaudill/coronadatascraper/blob/c64cf9c506b445e34a0c11776874b357ad94bf52/.editorconfig#L3

Happy to change it in both places if that's what we want, lmk.

scrapers/ESP/index.js
HerbCaudill: It's as close as we can get: Official data is published in PDF form, this repo just republishes it in CSV.

scrapers/ESP/index.js
HerbCaudill: :+1: done

package.json
camjc: Not sure where we landed. @lazd ?

lib/parse.js
HerbCaudill: Good point. 

In my case I'm getting `undefined` due to the way the scraped data is structured, so I've handled that within the scraper. 

I've changed parse.number so that it explicitly checks for `undefined` and throws a more informative error, rather than waiting until it tries to do `.replace` on an undefined value.

package.json
HerbCaudill: I see now that .editorconfig changed to 120 - I've updated the prettier config.

HerbCaudill:
@lazd is there anything else you need me to do to get this merged? No rush, just want to make sure the ball's not in my court.

package.json
```diff
@@ -86,5 +87,10 @@
       "eslint --fix",
       "git add"
     ]
+  },
+  "prettier": {
+    "printWidth": 120,
+    "semi": true,
+    "singleQuote": true
```
qgolsteyn: Let's keep this out of this PR

src/events/crawler/scrapers/ESP/index.js
```diff
@@ -1,38 +1,149 @@
...
-        }
-      }
+    const isDate = s => s.includes('/');
+
+    const parseDate = s => {
+      // Source data currently uses ddmmyyyy format for dates,
+      // but is (maybe?) planning to switch to yyyymmdd.
+      // See https://github.com/datadista/datasets/issues/14
+      // When/if this happens, replace contents of this function with:
+      // return LocalDate.parse(s)
```
qgolsteyn: I see this issue was closed, and the dates in the source have changed to YYYY-MM-DD

src/events/crawler/scrapers/ESP/index.js
```diff
@@ -1,38 +1,149 @@
...
+    const dates = Object.keys(sampleRow)
+      .filter(isDate)
+      .map(parseDate)
+      .sort();
+
+    const firstDate = dates[0];
+    const lastDate = dates[dates.length - 1];
+
+    // use the scrape date, or the latest available
+    const queryDate = scrapeDate || lastDate;
```
qgolsteyn: Prefer returning an empty array in case we do not have data for today

src/events/crawler/scrapers/ESP/index.js
HerbCaudill: 👍 done

src/events/crawler/scrapers/ESP/index.js
HerbCaudill: 👍 done

package.json
HerbCaudill: After merging in current master, this is no longer in the diff, since it was added in #303 .
https://github.com/lazd/coronadatascraper/pull/228/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

PR #229, opened 2020-03-21
PARKED: RKI parser for DEU county-level data

noleti:
Parser for an API that looks like an APP backend of the Robert Koch Institute in Germany/DEU, providing official county-level data (no time series). Many of the county locations are provided with a location by the coronadatascraper backend.


scrapers/DEU/rki.js
```diff
@@ -0,0 +1,42 @@
...
+
+const scraper = {
+  country: 'DEU',
+  timeseries: false,
+  url: 'https://services7.arcgis.com/mOBPykOjAyBO2ZKk/arcgis/rest/services/RKI_Landkreisdaten/FeatureServer/0/query?where=1%3D1&outFields=*&returnGeometry=false&outSR=4326&f=json',
+  type: 'json',
+  curators: [
+    {
+      name: 'Robert Koch Institute',
+      email: ''
```
camjc: Why empty string here?

scrapers/DEU/rki.js
```diff
@@ -0,0 +1,42 @@
...
+  ],
+  async scraper() {
+    const res = await fetch.json(this.url);
+
+    // Data structure
+    // {'features':[{'attributes':{'cases': INT, 'deaths': INT, 'BL': STRING, 'county': STRING, 'EWZ': INT}},..., {'attributes':{'cases': INT, 'deaths': INT, 'BL': STRING, 'county': STRING, 'EWZ': INT}}]}
+
+    const { features } = res; // a list of dictionaries, one for each county
+
+    return features.map(row => {
```
camjc: NaBD but can wrap the object in parens to avoid the explicit return

scrapers/DEU/rki.js
noleti: The API is not exactly advertised by RKI, so I was not sure which mail to set

scrapers/DEU/rki.js
noleti: Sorry, I don't follow the comment. Disclaimer: I'm not really fluent in .js, but I can hack together stuff with good examples from other scrapers. Please feel free to improve!

jgehrcke:
@camjc just to make this explicit here in this PR, too: needs to be coordinated with #196 which ist still work in progress.

scrapers/DEU/rki.js
camjc: Oh, just that

() = > ({foo:’foo’})
Is equivalent your your current
() => { return {foo:’foo’}}

Doesn’t matter, just shorter :)

camjc:
Sweet, #196 is merged. Lets get this one in then :)

noleti:
note to myself: move parser to `src/events/crawler/scrapers`. I will do this later today (and also update my branch) when I have time. Will ping you when I am done.

jgehrcke:
> Sweet, #196 is merged. Lets get this one in then :)

Huhu @camjc my most important question is: what's the rationale here? Can someone elaborate? :)

As we have DEU state-level scrapers: in my mental model we have two "conflicting" approaches here, or at least it's not at all obvious why they should go very well together until someone explains that well.

As someone who's spent a bit of time with the topic of COVID-19 case data in Germany I also question this specific data source.

I also would like to add that a quick "@jgehrcke WDYT?" can help wonders when one tries to collaborate on the same country :-), before writing code.

About a country-level scraper I would like to add here (because it might not be obvious): as part of https://github.com/lazd/coronadatascraper/pull/196 we discussed the question of adding a country-wide scraper on top of the state-level ones, discussed some pros and cons, and the different possible ways of doing so. We affter all concluded that we don't want to do it.
In the following commit message I had linked to the relevant discussion in Slack:
https://github.com/lazd/coronadatascraper/pull/196/commits/36d4197c0530dad59e9a4ad0fcc8b1d1bce39322 (I write this because country vs. state is similar to state vs. county).

About a county-level scraper for Germany I want to also say that in an earlier discussion tended to agree that we'd prefer to keep things state-level in this project.

In any case, I believe that points like this should be thoroughly discussed before considering to change gears.

Cheers!

noleti:
I rebased the branch now. Parser still seems to work. I originally made this PR as I planned to use the data downstream in another project. I have built my own downstream parser now, so feel free to use this straight-forward parser if you want, or reject the PR.

lazd:
@jgehrcke I do believe county-level is something we want to see. This should work fine in tandem with the state level scrapers, and if this code looks good, I say we merge it.

jgehrcke:
> @jgehrcke I do believe county-level is something we want to see.

OK, then let's work towards this!

> if this code looks good

For starters, let's look at the data itself :).

In terms of data, I believe the following points are important:

1) County-level and state-level data should use consistent data sources (case sum should ~match for each point in time along time series).

2) If we want to use _current_ data (not consume time series data, but build up our own time series) then we should use a _non-RKI_ data source (explanation below).

3) Instead of (2) I suggest consuming time series data provided by RKI (explanation below). This was technically challenging, but should be pretty easy now through data files newly added to https://github.com/jgehrcke/covid-19-germany-gae (after we learned about the importance of consulting the RKI for historical data, see below).

About (2) and (3): I have prepared this plot: https://covid19-germany.appspot.com/static/data-sources-comparison-2020-03-30.png

It indicates that
- for "now" (today) and the last 2-3 days, RKI data significantly underestimates the actual case count, whereas non-RKI sources are better-suited for an **estimate** of the current count. That pattern consistently appeared within the last 2 weeks (not shown by the plot, of course).
- for the actual history, only the RKI properly accounts for delay in the reporting chain (non-RKI sources are not well-suited for reflecting historical data).

One conclusion is that if we were to merge the PR as-is (which builds up time series from _current_ RKI data state) we would build up a curve (for every county) that is systematically and significantly lower than the actually known case count.

noleti:
I agree with @jgehrcke's analysis of issues with RKI reporting. RKI only recently (I saw it after my initial PR) started to provide information on the reporting delay and the way they handle it.

That being said, a quick analysis of the ([RKI App](https://experience.arcgis.com/experience/478220a4c454480e823b17327b2bf1d4)) shows that there are API calls such as 

`https://services7.arcgis.com/mOBPykOjAyBO2ZKk/arcgis/rest/services/Covid19_RKI_Sums/FeatureServer/0/query?f=json&where=(Meldedatum%3Etimestamp%20%272020-03-01%2022%3A59%3A59%27%20AND%20(Meldedatum%3Ctimestamp%20%272020-03-30%2022%3A00%3A00%27%20OR%20Meldedatum%20%3E%20timestamp%20%272020-03-31%2021%3A59%3A59%27))%20AND%20(IdLandkreis%3D%2709162%27)&returnGeometry=false&spatialRel=esriSpatialRelIntersects&outFields=ObjectId%2CSummeFall%2CMeldedatum&orderByFields=Meldedatum%20asc&resultOffset=0&resultRecordCount=2000&cacheHint=true`

or URL decoded

`https://services7.arcgis.com/mOBPykOjAyBO2ZKk/arcgis/rest/services/Covid19_RKI_Sums/FeatureServer/0/query?f=json&where=(Meldedatum>timestamp '2020-03-01 22:59:59' AND (Meldedatum<timestamp '2020-03-30 22:00:00' OR Meldedatum > timestamp '2020-03-31 21:59:59')) AND (IdLandkreis='09162')&returnGeometry=false&spatialRel=esriSpatialRelIntersects&outFields=ObjectId,SummeFall,Meldedatum&orderByFields=Meldedatum asc&resultOffset=0&resultRecordCount=2000&cacheHint=true`

That call seems to provide timeseries data for a county (in this case, Munich), with query parameters (meldedatum=report date, IdLandkreis=county ID). So the call returns cumulative data for munich=09162 from march 1st onwards. County names need to be translated to IDs using the API call I originally had in this PR. One would have query each county seperately, though. Cases and deaths also require one call each, it seems.

I agree that even with this new API call, data for recent days might still be updated in retrospective, which might clash with the way this project is handling cashing.

I don't plan on updating this PR due to lack of bandwidth, but I'm leaving this info here in case someone else wants to pick up. The API is fairly easy to parse, after all.

jgehrcke:
> The API is fairly easy to parse, after all.

Read this only now, sorry! For the record, I have been [using that ArcGIS API](https://github.com/jgehrcke/covid-19-germany-gae/blob/master/tools/build-rki-csvs.py) since March 27 to provide a daily updated build of https://github.com/jgehrcke/covid-19-germany-gae/blob/master/cases-rki-by-ags.csv as well as https://github.com/jgehrcke/covid-19-germany-gae/blob/master/deaths-rki-by-state.csv.

hyperknot:
@jgehrcke right now we are focusing our efforts into releasing a new architecture, which supports first level subdivisons internally only. US is an exception as it'll support counties.

You resourse is great and we can one day implement it to have full county support for Germany, especially if you help us find a GeoJSON or Shapefile for the counties with official population numbers.

jzohrab:
Hi all, I've reviewed this a few times but am not sure what to do, given the extensive conversation above.  Is this a valid scraper, and should we merge it?  Cheers! jz

jzohrab:
Closing due to lack of response.  Thank you!

PR #237, opened 2020-03-21
Map Legend

NateBaldwinDesign:
* Branched from the `rollup` branch*

Creates canvas displaying color scale used in map with annotation for the min and max values. Min and max are input from the data sources.

image


lazd:
This got hairy so I merged it manually, with your name as the author. Thanks!

PR #243, opened 2020-03-22
Country Level IDs

hyperknot:
This basically introduces an ID system for scrapers / locations. I added it to AUS scrapers.

So the basic idea is that anything which has an id gets the feature and the population in one line. Basically `id3:AU-VIC` maps to `id3/au/vic.geojson` which also contains the most up-to-date population data.

Now to keep everything working, my idea was to add id + scraperId + parentId to locations, so we have some duplicate data. This way it works both with the old and the new system, and we can gradually add ids to more and more scrapers.

I was thinking that in the future we don't need the country / state fields at all, but use the parentId to reconstruct the hierarchy. But this is just an idea, we don't even need to implement that.

To check it out, `git clone git@github.com:hyperknot/country-levels.git` into the data-sources dir for now.

You can try it out like `yarn start -i id0:AU` or `yarn start -i id3:AU-VIC`


hyperknot:
@lazd @camjc

src/events/crawler/scrapers/AUS/index.js
```diff
@@ -19,14 +32,21 @@ const scraper = {
...
       states.push({
-        state,
+        state: id,
+        id,
+        parentId: scraper.country,
         cases
       });
     });
     states.push(transform.sumData(states));
+    // console.log(states);
```
camjc: Remove before merge

src/events/crawler/tasks/findPopulations.js
```diff
@@ -67,7 +68,12 @@ const generatePopulations = async ({ locations, featureCollection, report, optio
 
   const populations = await readPopulationData(featureCollection);
 
-  function getPopulation(location) {
+  async function getPopulation(location) {
+    // getting population if id is present
```
camjc: Perhaps should be imperative “get”

src/events/crawler/scrapers/AUS/index.js
```diff
@@ -5,8 +5,21 @@ import * as transform from '../../lib/transform.js';
...
+  Tasmania: 'id3:AU-TAS',
+  Victoria: 'id3:AU-VIC',
+  'Western Australia': 'id3:AU-WA'
+};
+
 const scraper = {
-  country: 'AUS',
+  country: 'id0:AU',
+  scraperId: 'id0:AU',
+  id: 'id0:AU',
```
camjc: How does one know/set these ids and what is the difference between these three?

src/events/crawler/scrapers/AUS/index.js
```diff
@@ -5,8 +5,21 @@ import * as transform from '../../lib/transform.js';
 // Set county to this if you only have state data, but this isn't the entire state
 // const UNASSIGNED = '(unassigned)';
 
+const idMap = {
+  'Australian Capital Territory': 'id3:AU-ACT',
```
camjc: Not sure what these 0 and 3s are doing. Why not build on AUS and ACT as a concatenation for example? I think I’m not quite following the intent here, most likely my fault but would love to clarify.

lazd:
@qgolsteyn can you review this one? I am busy today but this is super important work we should let go stale.

hyperknot:
I'm updating the readme but in a few minutes https://github.com/hyperknot/country-levels should have some good explanation.

src/events/crawler/tasks/findPopulations.js
hyperknot: fixed

src/events/crawler/scrapers/AUS/index.js
hyperknot: 1. Getting the ids is easy, a contributor just looks up the right value on https://github.com/hyperknot/country-levels/blob/master/docs/country_list.md
2. Currently there is no different, it's a workaround to make the current system running as well as to use the new id system. It's quite verbose and not elegant though.

src/events/crawler/scrapers/AUS/index.js
hyperknot: It's just 100% mirroring the country, unit, subunit and state levels in Natural Earth dataset. Basically Natural Earth is the standard public source dataset for country shapes and it has this hierarchy.
If you look at https://github.com/hyperknot/country-levels/blob/master/docs/country_list.md it's basically Natural Earth's hierarchy mapped to id0, id1, id2 and id3.

hyperknot:
I've pushed a much better readme to https://github.com/hyperknot/country-levels.

This PR does two things:
1. Introduces id codes for the lookup, it works really well.
2. Adds those additional ids, of which I'm not sure about. I'll really need @lazd 's input on that. I'll clean that one up a bit.

hyperknot:
OK, I simplified this A LOT. Pretty much all those ids are now gone, I'm following the existing system 100%. Can you please re-review it? It should be much closer to the original system now.

All we need to do is to change the country/state strings to return the shortcode instead of the long state name. 

Basically that's it, it'll figure out that we are talking about a code and shortcut getFeature and getPopulation.

Now, there is one thing I couldn't debug, that is for some reason after the last "✏️  dist/ratings.json written" line, the app keeps waiting in there. I have no idea what could it be or how to fix it.

hyperknot:
Also added -i --id cli argument so we can "yarn start -i id3:AU-ACT" without the long string of -l.

src/events/crawler/lib/countryLevels.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+function cleanProperties(feature) {
+  const prop = feature.properties;
+  feature.properties = {
+    name: prop.name,
+    population: prop.population,
+    wikidata_id: prop.wikidata_id,
+    ne_id: prop.ne_id,
+    cl_id: prop.cl_id
```
camjc: Is there a good reason to have these as abbreviations? I'm guessing NE is natural earth, not sure what CL is. A casual reader would likely be lost here.

camjc:
I guess your CI is failing because of Victoria; I've got a fix for that in my https://github.com/lazd/coronadatascraper/pull/274 so perhaps that needs to go first 🤷‍♂

camjc:
@hyperknot let me know if I can give you a hand getting this ready for the people with merge access to review :)

hyperknot:
I've just finished rewriting country levels based on OSM + Wikidata. https://github.com/hyperknot/country-levels

I'll update this PR for finally getting iso1 and iso2 codes in this project.

hyperknot:
I'll submit a new PR based on current master.

PR #248, opened 2020-03-22
fix: codemod cleanups

camjc:
Cleanup 15 lint warns via 2 codemods.


qgolsteyn:
Thanks for this linting fix!

PR #249, opened 2020-03-22
[patch] add AUS NT + SA

camjc:
Pretty straightforward for once
- NT has a widget that has the case count
- SA has a dedicated page with a case count in a simple paragraph

### DIFF
--- a/src/events/crawler/scrapers/AUS/NT/index.js
+++ b/src/events/crawler/scrapers/AUS/NT/index.js
@@ -1,0 +1,17 @@
+import * as parse from '../../../lib/parse.js';
+import * as fetch from '../../../lib/fetch.js';
+
+const scraper = {
+  country: 'AUS',
+  state: 'Northern Territory',
+  url: 'https://coronavirus.nt.gov.au/',
+  type: 'table',
+  priority: 1,
+  async scraper() {
+    const $ = await fetch.page(this.url);
+    const $rowWithCases = $('.header-widget p:first-of-type');
+    return [{ state: scraper.state, cases: parse.number($rowWithCases.text()) }];
+  }
+};
+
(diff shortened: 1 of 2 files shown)


src/events/crawler/scrapers/AUS/NT/index.js
```diff
@@ -0,0 +1,17 @@
...
+const scraper = {
+  country: 'AUS',
+  state: 'Northern Territory',
+  url: 'https://coronavirus.nt.gov.au/',
+  type: 'table',
+  priority: 1,
+  async scraper() {
+    const $ = await fetch.page(this.url);
+    const $rowWithCases = $('.header-widget p:first-of-type');
+    return [{ state: scraper.state, cases: parse.number($rowWithCases.text()) }];
```
lazd: No need to return an array, but it's not hurting anything
```suggestion
    return { state: scraper.state, cases: parse.number($rowWithCases.text()) };
```

src/events/crawler/scrapers/AUS/NT/index.js
```diff
@@ -0,0 +1,17 @@
+import * as parse from '../../../lib/parse.js';
+import * as fetch from '../../../lib/fetch.js';
+
+const scraper = {
+  country: 'AUS',
+  state: 'Northern Territory',
+  url: 'https://coronavirus.nt.gov.au/',
+  type: 'table',
+  priority: 1,
```
lazd: Add yourself as a maintainer! I think the source URL is fine here, but could you also add `sources: [{ name: "Whatever this source is called" }]`. Same below

camjc:
Closed in favour of https://github.com/lazd/coronadatascraper/pull/274

PR #257, opened 2020-03-22
Fixed MS, USA scraper as page HTML changed; added deaths

chunder:


src/events/crawler/scrapers/USA/MS/index.js
```diff
@@ -141,6 +141,29 @@ const scraper = {
 
       counties.push(transform.sumData(counties));
 
+      return counties;
+    },
+    '2020-3-20': async function() {
```
lazd: Should this be `2020-3-23`, or does this issue also affect the 20th and 21st?

chunder:
There were no cases reported on 20 and 21. So I am assuming the breaking change happened then.

Paul

PR #265, opened 2020-03-22
feat: add scraper for HI.

Jord-Holt:


src/events/crawler/scrapers/USA/HI/index.js
```diff
@@ -0,0 +1,75 @@
...
+          At time of writing, this county has no cases and is not shown in the data table.
+          Very likely to change. Additionally, index for state deaths will also need to be modified
+          when/if final county is added to table (deaths are only listed for the state and not per county row).
+        */
+      if (index > 2 && index < 7) {
+        // This is to exclude extraneous empty rows in table html.
+
+        rowCellValueList = $tr.text().split('\n');
+
+        const county = rowCellValueList[1].trim();
```
qgolsteyn: Would it be possible to check if the county name is one of Hawaii's counties rather than relying on the index value? That way, the scraper will be more resilient to the addition of new islands.

src/events/crawler/scrapers/USA/HI/index.js
```diff
@@ -0,0 +1,75 @@
...
+      }
+
+      if (index === 9) {
+        rowCellValueList = $tr.text().split('\n');
+
+        stateDeaths = rowCellValueList[2].substr(0, rowCellValueList[2].indexOf(' ')).trim();
+      }
+    });
+
+    const aggregationRecord = transform.sumData(counties);
```
qgolsteyn: For missing counties, I would add `counties = geography.addEmptyRegions(counties, this._counties, 'county');`

src/events/crawler/scrapers/USA/HI/index.js
Jord-Holt: Great idea. I was trying to find a way to avoid using the index and I think this will do nicely. Ill get this changed shortly!

Jord-Holt:
@qgolsteyn @lazd, Changes have been submitted. Please take a look and let me know if there is anything else we need to address. Thanks!

PR #271, opened 2020-03-23
SC Scraper

rossspoon:
A selenium scraper to South Carolina


package.json
```diff
@@ -6,7 +6,7 @@
   "license": "BSD",
   "type": "commonjs",
   "engines": {
-    "node": ">=12.15.0 <13.0.0"
+    "node": ">=12.15.0 <13.12.0"
```
ryanblock: FYI: stable Node.js only, please, no 13+.

package.json
```diff
@@ -42,6 +42,7 @@
     "needle": "^2.3.3",
     "pdfreader": "^1.0.6",
     "puppeteer": "^2.1.1",
+    "selenium-webdriver": "^4.0.0-alpha.7",
```
camjc: I've had flakiness with selenium at the best of times. Do we as a project
a) want selenium when we already have puppeteer? and
b) want an alpha of selenium which is likely even more flaky than usual?

lazd: Yes, highly recommend we use Puppeteer -- we can't take on another large dep as we move to cloud functions.

src/events/crawler/scrapers/USA/SC/index.js
```diff
@@ -1,89 +1,69 @@
...
+      // kill the annoying header
+      await driver.executeScript("document.querySelector('header').remove()");
+
+      // Switch to iframe
+      let ifr = driver.wait(until.elementLocated(By.css('iframe')), 10000);
+      await driver.switchTo().frame(ifr);
+
+      // Switch to iframe
+      ifr = driver.wait(until.elementLocated(By.css('iframe')), 5000);
+      await driver.switchTo().frame(ifr);
```
lazd: Can we just get the URL of the iframe and go directly to it?

package.json
rossspoon: It works, that is it gets the right data, but it runs very slow when I set the browser to headless.  I'll look into puppeteer and see what I can come up with.

thanks for the input.

src/events/crawler/scrapers/USA/SC/index.js
rossspoon: I have considered that.  I'm worried that the iframes's url might be dynamically created.
Do you think that url will be stable?

I have tried some like:
- find the iframe and get the URL
- do a new driver.get() with the iframe's url.

rossspoon: I finally wised up and found the restful request that fetched the data for SC, only to find that I was beat to the punch.

qgolsteyn:
Closing in favour of existing scraper for SC.

PR #272, opened 2020-03-23
VA Scraper

aed3:


qgolsteyn:
Amazing work! Awesome you managed to get data out of Tableau!

qgolsteyn:
First, awesome work! Really cool that you managed to parse a Tableau PDF. I was about to merge and saw that you are using `pdf-parse`. I'm wondering if we could use the PDF scraper already implemented to avoid adding another dependency. Happy to help do the refactoring.

aed3:
Thanks! I tried using the original pdf scraper, but the output didn't give something I could reliably use since there was no way of knowing where a number ended. However, it seems like the output of the original pdf parser changed and I could probably use it now. Will get started on that.

qgolsteyn:
Alright! Feel free to ask questions here or ping me on Slack if you need anything!

aed3:
Ok, I fixed it and will make a new pull request now.

PR #280, opened 2020-03-23
feat: Use WayBack machine as fallback to missing cache

qgolsteyn:
If we try to access a resource that doesn't exit currently, or for which we don't have cache for, we try to fetch from the Wayback machine as an alternative. Still needs more testing.


lazd:
@qgolsteyn I already started this work in: https://github.com/lazd/coronadatascraper/tree/wayback

Can you review what I had done previously and make sure there isn't anything I did that you're missing?

camjc:
This feature would provide a lot of value. Let me know if theres anything I can do to help out.

lazd:
@qgolsteyn let me know if you found anything in my branch that's missing here, otherwise we should continue to hack on this branch!

lazd:
@qgolsteyn why close?

PR #284, opened 2020-03-23
Canonicalize misspelled county names in MS, USA

chunder:


lazd:
YES. This fixed the confusion we had earlier.

PR #285, opened 2020-03-23
Place cloropleth layer under label layer

cflann:
Keeps label layer from the style/template map visible over the chloropleth layer.
Screen Shot 2020-03-23 at 2 36 24 PM


lazd:
WHAT! I had no idea you could do this! Thank you!

PR #288, opened 2020-03-23
feat: IND

ArnavD74:
Location is IND
Able to scrape, but currently can not collect static population data from local csv


.gitignore
```diff
@@ -3,3 +3,5 @@ dist/
 node_modules
 .DS_Store
 .vscode
+coronadatascraper-cache/*
+coronadatascraper-cache
```
qgolsteyn: This should be left out, `coronadatascraper-cache` is the submodule we use to store the cache for this project.

qgolsteyn:
Alright, population works! But I still notice that the crawlere doesn't recognize that. Let me investigate.

.gitignore
ArnavD74: My bad, I had not meant to include it initially.

lazd:
@ArnavD74 great work! Wondering why Jharkhand is not on the map?

image

ArnavD74:
> @ArnavD74 great work! Wondering why Jharkhand is not on the map?
> 
> image

It seems to be on the map for me currently? Can you check again and report

PR #291, opened 2020-03-23
yarn dataQualityReporter

samirrayani:
This PR adds 3 dataset validation tests using our testing framework: `jest`. The tests run against the datasets located @ `dist/locations.json` and `dist/timeseries-byLocation.json`

To run this on your own, ensure you have the most up to date datasets in `dist` and then run:

```
> yarn dataQualityReporter
```

Currently, we are performing four tests (with room to grow!):

- A simple timezone validation ensuring our `locationObject` has the same metadata as our `timeseriesObject`
- A test to ensure that `deaths <= cases <= tested` (per issue https://github.com/lazd/coronadatascraper/issues/185)
- 2 tests to ensure that our case and death counts monotonically increase (partially taking care of this issue: https://github.com/lazd/coronadatascraper/issues/184)

These tests do not block deploys or commits and are used to help us identify datasets with issues so that we can retroactively handle.

The results of the tests are written to `/dist/data-quality.json` after every build. This will output the results of our data quality checks in a way that can be later visualized or presented so we can take some action!


tests/validationHelpers.js
```diff
@@ -0,0 +1,35 @@
+const locations = require('../dist/locations.json');
+const timeseries = require('../dist/timeseries-byLocation.json');
```
samirrayani: @lazd @qgolsteyn should I pull from `coronadatascraper.com/*.json` instead of from `dist/` ?

qgolsteyn: I would say there isn't much use in testing coronadatascraper.com/*.json as this data would have been released already.

package.json
```diff
@@ -13,6 +13,7 @@
     "dev": "npx rollup -c --watch",
     "build": "npx yarn start && npx yarn timeseries && npx yarn buildSite",
     "buildSite": "npx rollup -c",
+    "dataValidation": "jest tests/*validation.*",
```
qgolsteyn: I would suggest including the `dataValidation` as part of the `build` or `deploy` script to make sure it gets run.

tests/validationHelpers.js
samirrayani: so the idea was to run this on our existing data and not any data produced during a dev run of a scrape. Perhaps a better/clearer name for it would be `dataQualityChecker` ? I think the missing piece is a summary that can report the quality of each dataset we have based on these test criteria. Did you or @lazd have a different idea in mind?

package.json
samirrayani: we definitely need some sort of report that can be run during `build` or `deploy`! Going to look into `jest-reporting` options and will update this ticket with what I find

samirrayani:
@qgolsteyn ok I refactored this a tiny bit and built a custom jest-reporter that will now write to `/dist/data-quality.json` after every build. This will output the results of our data quality checks in a way that can be later visualized or presented so we can take some action! Let me know what you think!

chunder:
Thanks for creating this!

Another test I think we should do is run each scraper with a mocked `fetch` that returns 404 and then assert that the scraper doesn't return any results. We have scrapers that return zeros when they error. Returning 0 cases/deaths/etc is not the desired behavior if the source in inaccessible.

tests/dataQualityHelpers.js
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable global-require */
+let locations = [];
+let timeseries = [];
+try {
+  locations = require('../dist/locations.json');
```
camjc: These will be generated dynamically, so might want to chuck some eslint ignores in here.

qgolsteyn:
Closing in favour of #397

samirrayani:
👍

PR #293, opened 2020-03-24
Ar

chunder:
Added county level scraper for AR, USA.


lazd:
@chunder looks like we're good on AR, can this be closed?

chunder:
Yup

PR #297, opened 2020-03-24
fix: broken CA scrapers

chunder:


qgolsteyn:
Failing CI makes sense, we added a failing scraper. We can safely ignore it.

PR #298, opened 2020-03-24
fix: VA County Names

aed3:
The names of some Virginia counties end with "city" in the data base I pulled from while they are "County" in the geoJSON file


src/events/crawler/scrapers/USA/VA/index.js
```diff
@@ -126,25 +125,25 @@ const scraper = {
...
     'Tazewell County',
-    'Virginia Beach city',
+    'Virginia Beach County',
     'Warren County',
     'Washington County',
-    'Waynesboro city',
+    'Waynesboro County',
     'Westmoreland County',
-    'Williamsburg city',
-    'Winchester city',
```
lazd: Ok, so these are actually counties, not cities? Or are they city-counties (same name for both city and county)?

aed3: Some of them are city-counties while others are different counties and cities with the same name.

lazd:
@aed3 can you please separate the VA counties from the NE scraper and send the NE scraper in a separate PR?

aed3:
> @aed3 can you please separate the VA counties from the NE scraper and send the NE scraper in a separate PR?

Yes

PR #301, opened 2020-03-24
docs: explain what data we need

camjc:
Hopefully we can use this to lobby governments for better sources.

Screen Shot 2020-03-24 at 8 40 26 pm

Screen Shot 2020-03-24 at 8 40 35 pm

### DIFF
--- a/site/example.csv
+++ b/site/example.csv
@@ -1,0 +1,3 @@
+datetime,             county,        state,  country,  cases,  deaths,  recovered,  tested
+2020-03-24T03:54:24Z, Kings County,  CA,     USA,      10,     1,       3,          100
+2020-03-23T03:54:24Z, Kings County,  CA,     USA,      8,      0,       2,          80
--- a/site/example.csv
+++ b/site/example.csv
@@ -1,3 +1,3 @@
-datetime,             county,        state,  country,  cases,  deaths,  recovered,  tested
-2020-03-24T03:54:24Z, Kings County,  CA,     USA,      10,     1,       3,          100
-2020-03-23T03:54:24Z, Kings County,  CA,     USA,      8,      0,       2,          80
+datetime,   county,        state,  country,  cases,  deaths,  recovered,  tested
+2020-03-24, Kings County,  CA,     USA,      10,     1,       3,          100
+2020-03-23, Kings County,  CA,     USA,      8,      0,       2,          80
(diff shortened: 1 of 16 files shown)


chunder:
Did you mean to include the changes to the linter config and maintainers?

camjc:
Hey, yeah sorry a bit unrelated. Happy to split those out this arvo (I figured they were trivial enough to chuck in here, should have explicitly mentioned, sorry!)

chunder:
Ok, just wanted to check they weren't included by mistake.

camjc:
@chunder Ideally we'd want timestamps too right? Some places are providing data multiple times a day

camjc:
Removed any linting stuff from here; it shouldn't be here and is conflicting anyway.

camjc:
@chunder What are your thoughts on adding time?
- If we don't want to do it we should rename column header to `date`
- If we do want to do it we should revert that commit

chunder:
I feel like this page should be asking for pragmatic, simple things from agencies. If they're sophisticated enough to be comfortable with UTC timestamps, they're probably not looking to us for advice on how to structure their CSVs. They probably are publishing clean data already.

If someone's being directed to read this page, it's because they are making poor decisions like publishing the data as an image of a map, etc. Given that target audience, I feel like we should be offering simple advice on how to produce *better* data. Not perfect data.

So all that said, a simple date column is probably a big win.

site/example.csv
```diff
@@ -0,0 +1,3 @@
+datetime,   county,        state,  country,  cases,  deaths,  recovered,  tested
```
camjc:
```suggestion
date,       county,        state,  country,  cases,  deaths,  recovered,  tested
```

ryanblock:
Personally, I'd lead with the actual **data** we care more about most, as that's the most important (for obvious reasons).

I'd then advise on presentation layer (which, admittedly, is also quite important!) and other factors.

camjc:
@ryanblock Can you chuck a commit in with what your suggestion is?

ryanblock:
In the interest of not PRing your PR (yo dawg), here's a blob:
```
First, please have a look at our <a href="example.csv" download class="spectrum-Link">example format</a>.<br/>

As the data itself is most important, please publish your total case count, deaths, recovered, and total tests administered.

If you cannot publish in JSON or CSV, at a minimum, please include a HTML <code>&lt;table&gt;</code> with one row per locality at the most granular level you have.<br/>

Have a column for the name of this locality, and a column for each additional data point.<br/>

The best published sources include timestamps, allowing citizens and researchers alike better understand the data you have, over time, within your specific geographic region(s).<br/>
```

camjc:
@ryanblock LGTM, thanks dawg ;)

qgolsteyn:
Let's start with that and adjust if necessary.

PR #308, opened 2020-03-24
Add .catch to fetch (new KS, USA scraper)

ndom91:
So that the entire process doesn't get killed

### DIFF
--- a/src/events/crawler/scrapers/USA/KS/index.js
+++ b/src/events/crawler/scrapers/USA/KS/index.js
@@ -136,7 +136,7 @@ const scraper = {
       const datePart = datetime.getMonthDYYYY(date);
       this.url = `${this._baseUrl}COVID-19_${datePart}_.pdf`;
 
-      const body = await fetch.pdf(this.url);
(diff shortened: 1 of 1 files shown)


lazd:
@qgolsteyn was looking at fixing this for everything using `domain`. We already catch outside of this, so it shouldn't hose the whole process. Right now, only socket errors can take it out, for some reason...

ndom91:
@lazd for some reason this does hose the whole process though unfortunately:

image

lazd:
Yeah, that's exactly the error we were talking about... Seems like throwing the catch on the promise is a potential solution then too. @qgolsteyn?

qgolsteyn:
Yes, I just tested this on my machine. We essentially need to catch the error, and avoid throwing again or it will crash the whole process.

qgolsteyn:
We should change the `fetch` function directly. That way, all scrapers will get the fix. Will close this PR once I have a global fix.

qgolsteyn:
Closing in favour of #321

PR #318, opened 2020-03-24
patch: Kern and Solano, CA, USA

chunder:


src/events/crawler/scrapers/USA/CA/solano-county.js
```diff
@@ -28,6 +28,9 @@ const scraper = {
       const deaths = parse.number(lines[2].split(':')[1]);
 
       return { cases, deaths };
+    },
+    '2020-3-24': async function() {
+      throw new Error('Solano County, CA now uses a PDF');
```
qgolsteyn: Can we scrape this PDF? Or does it contain an image?

chunder: Probably. But it seemed not worth the effort for a single county in CA where when we are using Harriets manual data anyway.

PR #328, opened 2020-03-25
Fixed broken AK, USA scraper

chunder:


src/events/crawler/scrapers/USA/AK/index.js
```diff
@@ -60,19 +60,21 @@ const scraper = {
   async scraper() {
     const counties = [];
     const $ = await fetch.page(this.url);
-    const $table = $('th:contains("Region")').closest('table');
+    const $table = $('td:contains("Gulf Coast")').closest('table');
```
qgolsteyn: Will this work for earlier dates?

chunder: Yes

PR #329, opened 2020-03-25
Allow up to 5 retries if fetch returns a recoverable error; Made VA t…

chunder:
…olerant to individual counties erroring


chunder:
@lazd you want to review?

PR #331, opened 2020-03-25
Fixed CT, USA scraper

chunder:


src/events/crawler/scrapers/USA/CT/index.js
```diff
@@ -99,8 +99,20 @@ const scraper = {
         if (!Number.isNaN(parse.number(rows[i][0]))) continue;
 
         const countyName = geography.addCounty(rows[i][0]);
-        const cases = parse.number(rows[i][1]);
-        const deaths = parse.number(rows[i][3]);
+
+        let cases = null;
+        let deaths = null;
```
qgolsteyn: Should set `cases` or `deaths` to undefined if we don't have a value for them?

chunder: The code will either give them a value or throw an error.

chunder: Also, the linter won't let me initialize them to undefined!

chunder:
Ok, no more null.

PR #333, opened 2020-03-25
fix: AUS Victoria

camjc:
For Victoria, AUS, the 03-25 press release has the word `is` instead of `to` before the number of cases, which breaks the regex

### DIFF
--- a/src/events/crawler/scrapers/AUS/VIC/index.js
+++ b/src/events/crawler/scrapers/AUS/VIC/index.js
@@ -22,7 +22,7 @@ const scraper = {
     const currentArticleUrl = $anchor.attr('href');
     const $currentArticlePage = await fetch.page(`https://www.dhhs.vic.gov.au${currentArticleUrl}`);
     const paragraph = $currentArticlePage('.page-content p:first-of-type').text();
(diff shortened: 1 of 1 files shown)


src/events/crawler/scrapers/AUS/VIC/index.js
```diff
@@ -22,7 +22,7 @@ const scraper = {
     const currentArticleUrl = $anchor.attr('href');
     const $currentArticlePage = await fetch.page(`https://www.dhhs.vic.gov.au${currentArticleUrl}`);
     const paragraph = $currentArticlePage('.page-content p:first-of-type').text();
-    const { casesString } = paragraph.match(/cases in Victoria to (?<casesString>\d+)./).groups;
+    const { casesString } = paragraph.match(/cases in Victoria .* (?<casesString>\d+)./).groups;
```
qgolsteyn:
```suggestion
    const { casesString } = paragraph.match(/cases in Victoria \w* (?<casesString>\d+)./).groups;
```

I found `.*` to skip the case count and go to the next number (55)

camjc: Even better. Cheers Q

PR #337, opened 2020-03-25
added fixed source for 3/24 (cases and deaths)

cmcjacob:
this is my first PR so apologies if something is off, Texas changed sources today so data now pulled from ARCGIS.  Only cases and deaths are reporting from the new source.


cmcjacob:
fixed

Issue #307, opened 2020-03-24
KS should not return all 0s if it gets an error fetching the source PDF


chunder:
It should return no results, not 0s for all counties.

```
      const body = await fetch.pdf(this.url);
      if (body === null) {
        // throw new Error(`No data for ${date}`);
        return geography.addEmptyRegions([], this._counties, 'county');
      }
```

chunder:
@paulboal ^^

paulboal:
Question @chunder - when there's no content, should the parser function just `return [];` or something else?  I'll get this fixed.

chunder:
Good question. @lazd ?

paulboal:
Looks like this was already resolved.  This can be closed.

lazd:
Sounds like this is related to #374, adding a note here.

PR #340, opened 2020-03-25
fix: #307 KS should not return 0 when no data

paulboal:
This is a fix for issue #307 - returning empty list instead of a list with all counties populated with 0s.


paulboal:
Looks like this was already addressed with a new throw.

PR #342, opened 2020-03-25
use HHS level data from AUS QLD

camjc:
AUS QLD have added a table, woohoo! Now they're reporting total confirmed to date instead of just new additions.

TODO
- [ ] Fix the total for QLD, it goes to 794 for some reason (double the total). Not sure why its doing this, hopefully someone can enlighten me.
- [ ] Figure out what is supposed to happen for "overseas" as a reporting area 😬 .
- [ ] How to map these HHS regions onto the geo map

### DIFF
--- a/src/events/crawler/scrapers/AUS/QLD/index.js
+++ b/src/events/crawler/scrapers/AUS/QLD/index.js
@@ -1,6 +1,14 @@
-import * as parse from '../../../lib/parse.js';
 import * as fetch from '../../../lib/fetch.js';
+import * as parse from '../../../lib/parse.js';
+import * as transform from '../../../lib/transform.js';
 import maintainers from '../../../lib/maintainers.js';
+
+async function getCurrentArticlePage(listUrl) {
+  const $ = await fetch.page(listUrl);
+  const anchors = $('#content h3:first-of-type > a');
+  const currentArticleUrl = anchors[0].attribs.href;
+  return fetch.page(currentArticleUrl);
+}
 
 const scraper = {
   country: 'AUS',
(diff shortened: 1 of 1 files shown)


src/events/crawler/scrapers/AUS/QLD/index.js
```diff
@@ -16,18 +24,35 @@ const scraper = {
...
+      const HHSs = []; // See https://www.health.qld.gov.au/maps
+      const $ = await getCurrentArticlePage(this.url);
+      const $table = $('#content table');
+      const $trs = $table.find('tbody > tr:not(:last-child)');
+      $trs.each((index, tr) => {
+        const $tr = $(tr);
+        const cases = parse.number($tr.find('td:last-child').text());
+        const county = parse.string($tr.find('td:first-child').text());
+        HHSs.push({
+          county,
```
lazd: Hmm, the previous scraper returned states, now this one has counties? Trying to wrap my head around the geography.

camjc: Previous scraper was a state. Now it has subdivided areas called HHS. Can see the map in the comment.
Potentially I’d marked it wrong either before or after, but the idea is going from one state (Queensland) to a dozen or so sub-areas (which I figure aligns to USA counties).

Annoyingly one of the sub areas is “overseas” so not sure how to handle that either.

lazd:
@camjc the total is double because it counts summed data and individual regions data

camjc:
@lazd so this is intended behaviour or I should fix something?

lazd:
@camjc intended behavior -- those totals are double counted.

lazd:
@camjc you're seeing that in the build log output, not data, right?

PR #345, opened 2020-03-25
tests: deterministic scraper tests

HerbCaudill:
Going forward we're going to need to refactor some things in ways that could break existing scrapers if we're not careful. To allow us to move faster with this work, I'd like to get at least some of the scrapers under test. 

This PR introduces a way to easily test that a scraper, given a specific response, returns a given dataset. 

## How it works

To add test coverage for a scraper, you only need to provide test assets; no new tests need to be added.

- Add a `tests` folder to the scraper folder, e.g. `scrapers/FRA/tests` or `scrapers/USA/AK/tests` 
- Add a sample response from the target URL. The filename should be the URL, without the
  `http(s)://` prefix, and with all non-alphanumeric characters replaced with an underscore `_`. The
  file extension should match the format of the contents (`html`, `csv`, `json`, etc). Example: 
    - URL: https://raw.githubusercontent.com/opencovid19-fr/data/master/dist/chiffres-cles.csv
    - File name: raw_githubusercontent_com_opencovid19_fr_data_master_dist_chiffres_cles.csv

- Add a file named `expected.json` containing the array of values that the scraper is expected to
  return. (Leave out any geojson `features` properties.)
- For sources that have a time series, the `expected.json` file represents the latest result in the
  sample response provided. You can additionally test the return value for a specific date by adding
  a file with the name `expected.YYYY-MM-DD.json`; for example, `expected.2020-03-16.json`. 

      📁 USA
        📁 AK
          📄 index.js 🡐 scraper
          📁 tests
            📄 dhss_alaska_gov_dph_Epi_id_Pages_COVID_19_monitoring.html 🡐 sample response
            📄 expected.json 🡐 expected result
      ...
      📁 FRA
        📄 index.js 🡐 scraper
        📁 tests
          📄 raw_githubusercontent_com_covid19_fr_data_chiffres_cles.csv 🡐 sample response
          📄 expected.json 🡐 expected result for most recent date in sample
          📄 expected.2020-03-16.json 🡐 expected result for March 16, 2020

## Changes

- Add test assets to USA/AK, USA/AL, and FRA.
- Refactor `fetch.js`: In order to be able to mock `fetch.fetch()` independently of `fetch.csv()`, `fetch.json()`, etc., I've pulled `fetch.fetch()` into its own module, called `get`. `fetch.js` then imports it and exposes it as `fetch.fetch()` so no existing code has to change. 
- Add `scrapers.all.test.js`, which finds any scrapers with test assets and tests them. 

A couple of incidental changes:
- The FRA scraper assumed that the latest changes are for the current date; presumably that's always been the case when it's been run, but can't be assumed and also makes it impossible to test deterministically. I've fixed the logic so it matches the other timeseries scrapers. 
- Added a shared `join` function to `lib` normalizes to forward slashes, since the issue in #251 comes up in a couple of other places. 

## Possible next steps

I've created the existing test assets manually, but it would be pretty easy to script generating assets for existing scrapers.


.eslintrc.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": ["airbnb-base", "plugin:jest/recommended", "plugin:jest/style", "plugin:prettier/recommended"],
-  "ignorePatterns": ["dist/"],
+  "ignorePatterns": ["dist/", "src/events/crawler/scrapers/**/tests/"],
```
HerbCaudill: don't lint test assets

package.json
```diff
@@ -90,5 +90,11 @@
     "printWidth": 120,
     "semi": true,
     "singleQuote": true
+  },
+  "jest": {
+    "rootDir": "src",
+    "testMatch": [
+      "**/?(*.)test.js"
+    ]
```
HerbCaudill: narrows the search space for test files, which should make jest a little faster to start up.

src/events/crawler/lib/fetch.js
```diff
@@ -1,107 +1,20 @@
...
-import puppeteer from 'puppeteer';
 import { PdfReader } from 'pdfreader';
-
-import * as datetime from './datetime.js';
+import puppeteer from 'puppeteer';
 import * as caching from './caching.js';
+import * as datetime from './datetime.js';
+import { get } from './get.js';
+
+export const fetch = get;
```
qgolsteyn: NIT: a small comment to explain why we have to do this.

HerbCaudill: 👍

chunder:
Great initiative. A detailed review of this js is above my pay grade. I hope to crib some of this structure for other tests I have in mind.

PR #348, opened 2020-03-25
Fixed DeKalb, GA and VA County Names

aed3:
The original VA County Names PR was overwritten

### DIFF
--- a/src/events/crawler/scrapers/USA/GA/index.js
+++ b/src/events/crawler/scrapers/USA/GA/index.js
@@ -180,7 +180,8 @@ const scraper = {
     const $trs = $table.find('tbody > tr');
     $trs.each((index, tr) => {
       const $tr = $(tr);
-      let county = geography.addCounty(parse.string($tr.find('td:first-child').text()));
+      const name = $tr.find('td:first-child').text();
+      let county = geography.addCounty(parse.string(name.replace('Dekalb', 'DeKalb')));
       const cases = parse.number($tr.find('td:last-child').text());
       if (county === 'Unknown County') {
         county = UNASSIGNED;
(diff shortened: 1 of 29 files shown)


src/events/crawler/scrapers/USA/GA/index.js
```diff
@@ -180,7 +180,8 @@ const scraper = {
     const $trs = $table.find('tbody > tr');
     $trs.each((index, tr) => {
       const $tr = $(tr);
-      let county = geography.addCounty(parse.string($tr.find('td:first-child').text()));
+      const name = $tr.find('td:first-child').text();
+      let county = geography.addCounty(parse.string(name.replace('Dekalb', 'DeKalb')));
```
chunder: Looks like a nice tactical fix for this.

aed3: Personally, I feel the best way to fix it would be to ignore case for every county name because I have a suspension this isn't the only place capitalization will cause problems. But yeah, this works for now.

src/events/crawler/scrapers/USA/VA/index.js
```diff
@@ -126,25 +125,25 @@ const scraper = {
...
+    'Staunton County',
+    'Suffolk County',
     'Surry County',
     'Sussex County',
     'Tazewell County',
-    'Virginia Beach city',
+    'Virginia Beach County',
     'Warren County',
     'Washington County',
-    'Waynesboro city',
```
chunder: Do we need to canonicalize these names in the `coronavirus-data-sources` too? For example, Waynseboro appears as "city" in the geojson but "County" in the population csv.

```
$ grep Waynesboro */*
geojson/usa-counties.json:        "name": "Waynesboro County, VA"
population/population-usa-counties.csv:"Waynesboro city, VA",22628
```

aed3: Yes, already put in a PR for that.

src/events/crawler/scrapers/USA/GA/index.js
chunder: Agreed. See this issue for more context:

https://github.com/lazd/coronadatascraper/issues/314

lazd:
#371

PR #351, opened 2020-03-25
Add ESP test assets

HerbCaudill:


HerbCaudill:
This doesn't affect any running code. Only review necessary is that test suite passes, which is provided by CI. (Side note - maybe it would be worthwhile to have an `assets-only` label or something, so things like this can get fast-tracked?)

PR #352, opened 2020-03-25
Issue 263: remove maintainers from dist csv

jzohrab:
Fix for https://github.com/lazd/coronadatascraper/issues/263.

Verified manually locally with `yarn start --location "Ventura County, CA, USA"`, and the rest of the columns were unchanged:

Old headings: `city,county,state,country,cases,deaths,recovered,tested,active,population,lat,long,url,maintainers,rating,tz,featureId`

New headings: `city,county,state,country,cases,deaths,recovered,tested,active,population,lat,long,url,rating,tz,featureId`

I don't see a way to add a regression test for this, but if there is, let me know and I'll update the PR.  Cheers!

### DIFF
--- a/src/events/crawler/lib/stringify.js
+++ b/src/events/crawler/lib/stringify.js
@@ -29,8 +29,9 @@ export const csvForDay = function(data) {
     }
   }
 
-  // Drop coordinates
-  columns = columns.filter(column => column !== 'coordinates');
+  // Drop unwanted columns.
+  const unwanted = ['coordinates', 'maintainers'];
+  columns = columns.filter(column => !unwanted.includes(column));
 
   // Turn data into arrays
   const csvData = [columns];


src/events/crawler/lib/stringify.js
```diff
@@ -29,8 +29,9 @@ export const csvForDay = function(data) {
     }
   }
 
-  // Drop coordinates
-  columns = columns.filter(column => column !== 'coordinates');
+  // Drop unwanted columns.
+  const unwanted = ['coordinates', 'maintainers'];
+  columns = columns.filter(column => !unwanted.includes(column));
```
lazd: What about curators/sources? They're not coming through?

jzohrab: According to crawler/lib/stringify.js, no.  The first columns (up to `url`) are included by default, and then the code adds extra columns according to what's in the data being written.

jzohrab: Note: the headings shown could just be due to the way that I ran my manual test, using only one county.  I didn't generate it using all locations.

jzohrab:
Code subsequently added to `src/events/processor/write-data/stringify.js` in another PR ... wasted effort! :-( :-)

lazd:
Totally my bad. I just fixed this manually without checking if there was a PR open. Sorry!

PR #353, opened 2020-03-25
Fix DeKalb County, GA missing data

rstallings:
DeKalb County, GA is incorrectly capitalized on the state covide site.
This causes the data to be missed.  This change special cases DeKalb to
fix the feed data.


rstallings:
Looks like it is already addressed in #348.   Closing.

PR #360, opened 2020-03-25
Temporary fix for Illinois scraper.

alexmill:
Illinois is not using a consistent format for their filenames, so they have needed to be updated manually for each of the past 3-4 days. This fixes the URL for today's scrape (3/25). See related issue https://github.com/lazd/coronadatascraper/issues/290


alexmill:
This is just a temp fix. I've proposed a more permanent solution here https://github.com/lazd/coronadatascraper/issues/290#issuecomment-603634951; my JS dev is a bit out of date, but I'll see if I can make some progress on it.

PR #366, opened 2020-03-26
Fixes Chicago/Cook county issue and fetches source URL daily

alexmill:
This should fix https://github.com/lazd/coronadatascraper/issues/290 and https://github.com/lazd/coronadatascraper/issues/339. 

Happy to take suggestions if there are issues with my implmentation.

### DIFF
--- a/src/events/crawler/scrapers/USA/IL/index.js
+++ b/src/events/crawler/scrapers/USA/IL/index.js
@@ -20,28 +20,19 @@ const scraper = {
     
     if (datetime.dateIsBefore(date, '2020-03-24')) {
       this.url = `${this._baseUrl}sites/default/files/COVID19/COVID19CountyResults.json`;
-    
     } else if (datetime.dateIsBefore(date, '2020-03-25')) {
       this.url = `${this._baseUrl}sites/default/files/COVID19/COVID19CountyResults20200323.json`;
-    
     } else if (datetime.dateIsBefore(date, '2020-03-26')) {
       this.url = 'http://www.dph.illinois.gov/sitefiles/COVIDTestResults.json';
     } else {
-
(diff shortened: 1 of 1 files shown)


lazd:
This looks complicated and I imagine we'll have to refactor it later to user separate scraper functions for each date, but if it works, that's great.

Did you test a timeseries for previous dates?

`yarn timeseries -d 2020-3-20 -l "IL, USA"`

alexmill:
Yeah, I've run `yarn timeseries -d 2020-03-19 -e 2020-03-25 -l 'IL, USA'` which seems to work okay. There are definitely some issues with the historical data to be worked out. But this should solve the foward-looking problem of making sure the current JSON file fetched each day. 

I will keep an eye on it to make sure it works for the next few days.

alexmill:
Actually, now that I think of it, this problem may be more complicated than I thought. We need a way of saving the URL associated with each date, since it may change daily. 

If nothing else, the code I used to scrape the source page and extract that URL can be built upon. But I'm not sure how to do the above.

lazd:
@chunder what's up with this one? Was it fixed in #381 ?

alexmill:
https://github.com/lazd/coronadatascraper/commit/b7a6235e75f4e46d17d96304289fd860c05662eb solved https://github.com/lazd/coronadatascraper/issues/339 but not https://github.com/lazd/coronadatascraper/issues/290. I'll need to resolve my code with that commit so that it just addresses https://github.com/lazd/coronadatascraper/issues/290.

alexmill:
There are issues with the historical data here that I'm not sure how to solve. I'm going to close this PR and just describe my proposed fix in https://github.com/lazd/coronadatascraper/issues/290 in case someone with more knowledge wants to fix this problem permanently.

PR #367, opened 2020-03-26
Added graphing of time series data when location is clicked 

aed3:


camjc:
Excited to get something like this off the ground. @lazd was looking at something like this too. Whats the plan and how can I help out?

camjc:
Probably need to revert the cache? Seems like this pr doesn’t mean to touch it?

aed3:
> Probably need to revert the cache? Seems like this pr doesn’t mean to touch it?

You're right, but I'm not sure how to do that.

aed3:
But the feature works 100% now, on mobile and everything. I can save the work for COVID Atlas, but don't think it would hurt to add to the current website @lazd

lazd:
Yeah this looks good, we can merge!

site/index.html
```diff
@@ -194,11 +196,17 @@ <h2 class="spectrum-Heading spectrum-Heading--M" id="spec">What does a good sour
...
+            <div id="graph-container">
+              <canvas id="graph"></canvas>
             </div>
-            <div id="map"></div>
+            <span onclick="document.getElementById('graph-elements').style.visibility='hidden'" id="graph-close">×</span>
+          </div>
+          <div class="cds-Map-controls">
           </div>
+          <div id="map"></div>
         </div>
```
lazd, line 210: Unclosed `<div>` broke viewer #820 :'(

PR #376, opened 2020-03-26
[patch] fix AUS ACT

camjc:
I have no clue why this works but it seems to. Can anyone enlighten me? It scraped 0 last day.


camjc:
How to fix the erroneous 0 from history?

qgolsteyn:
If you mean yesterday's timeseries, it should be removed the next time we update the site as long as the scraper works

camjc:
@qgolsteyn are we able to merge?

PR #380, opened 2020-03-26
feat: Roll Chicago data into Cook County

jeffchenoweth:
Chicago not in summary totals for Illinois; should be in Cook County


jeffchenoweth:
I screwed up backwards compatibility...fixing now.

PR #382, opened 2020-03-26
getArcGISCSVURL without the server name

joshuaellinger:
Gets the full manifest and finds the data source based on
widget name.  From the data source, you can get the server.

** This code has not been run.  Please validate if you merge it **

-----
The key insight is that you can get a full manifest by looking at /data.
From there, you have to find the widget and get the data source,
then pull the meta-data for the source which contains the services
URL.

It is not 100% clear if we'd be better searching on name or caption.
to find the widget.  Caption is the HTML title for widget.  Name is
internal and not present in rendered UI.  I choose name because
Texas uses an image.


src/events/crawler/lib/fetch.js
```diff
@@ -210,3 +210,49 @@ export const getArcGISCSVURL = async function(serverNumber, dashboardId, layerNa
...
+ */
+export const getArcGISCSVURL_NEW = async function(dashboardId, widgetName) {
+
+  // 1. get the full manifest
+  const dashboardManifest  = await json(`https://maps.arcgis.com/sharing/rest/content/items/${dashboardId}/data?f=json`);
+
+  const { orgId } = dashboardManifest;
+  
+  // 2. find the datasourceID for the target 
+  const datasourceId = findArcGISFeatureDataSourceIDByName(dashboardManifest, widgetName)
```
lazd: @joshuaellinger is this function missing?

src/events/crawler/lib/fetch.js
```diff
@@ -210,3 +210,49 @@ export const getArcGISCSVURL = async function(serverNumber, dashboardId, layerNa
...
+ *
+ * To get the widgetName, load the manifest, reformat it, and search for a widget that has a caption
+ * that matches the title of table shown in the UI.  The caption is HTML but some places put images 
+ * instead of text so we should the name when scraping.  For example,
+ *    https://maps.arcgis.com/sharing/rest/content/items/ec4bffd48f7e495182226eee7962b422/data?f=json
+ * has a widget named 'COVID-19 Cases by County' which has a datasourceId = 24f4fcf164ad4b4280f08c8939dd5dc7
+ *
+ *
+ */
+export const getArcGISCSVURL_NEW = async function(dashboardId, widgetName) {
```
lazd: Let's call it `getArcGISCSVURLByWidgetName`
```suggestion
export const getArcGISCSVURLByWidgetName = async function(dashboardId, widgetName) {
```

joshuaellinger: Lost the function when I forked the repo.  I was originally working on a local branch.

joshuaellinger:
Closing the request and replacing it.

joshuaellinger:
Okay -- I fixed the lint errors but hit a conflict on README.md caused by adding a note that you need to install node.

I have to admit that I really don't like the modern javascript world.  It looks cleaner thanks to async but it is still such a weak foundation.  Typescript makes it bearable.  I'm going to head back to Python.  Can you finish it from here?

lazd:
Fixed the merge issues, but when I pushed, github closed the PR. Odd.

PR #393, opened 2020-03-26
Introducing new ISO date utilities

HerbCaudill:
This PR introduces the beginnings of a new utility module for working with dates, times, and timezones. 

At this point there's not much here, but wanted to introduce this as a foundation and then work at it incrementally. 

### The API so far

`today` gives the current date, in the local machine's timezone or in another.

```js
// suppose current time is 23:45 UTC & date is March 16, 2020

today.here(); // '2020-03-16'
today.at('UTC'); // '2020-03-16'
today.at('Australia/Sydney'); // '2020-03-17'
```

Similarly, `now` gives the current date+time.

```js
now.here(); // '2020-03-16T23:45'
now.at('UTC'); // '2020-03-16T23:45'
now.at('Australia/Sydney'); // '2020-03-17T10:45'
```

`parse`

```js
parse(new Date('2020-03-16')); // '2020-03-16'

parse('2020-03-16'); // '2020-03-16'
parse('2020-03-16T23:45:00Z'); // '2020-03-16'
parse('2020-3-1'); // '2020-03-01'
```

### Principles

- The intention is to eliminate any usage of the Javascript `Date` object anywhere in the codebase. (See [here](https://jj09.net/javascript-date-a-bad-part/) and [here](https://codeofmatt.com/javascript-date-type-is-horribly-broken/) for reasons why the built-in `Date` object is broken.) Code for the crawler, processor, and all scrapers will just work with strings in ISO format, using this module for any parsing, comparing, and timezone calculations. 

- The intention is to incrementally replace calls to the current `datetime` methods with this, until that module can be removed. 

- This module uses js-joda internally; no other code should depend on js-joda, so that the underlying date/time/timezone library could be changed if needed. 

### Testing across multiple timezones

Jest runs all tests assuming the timezone of the running operating system. This means that it's possible to write passing tests that then fail during Daylight Savings Time, or that fail when run by a developer in another timezone. 

Turns out it's [surprisingly](https://github.com/jest-community/vscode-jest/issues/153) [difficult](https://stackoverflow.com/questions/56261381/how-do-i-set-a-timezone-in-my-jest-config) to mock the timezone in node.js in a cross-platform way; but I _think_ I've cracked that nut. I've added a node script, `scripts/testTimezones.js` that detects the OS, sets the timezone accordingly, and then runs jest in a child process. 

I've tested this test in Windows and WSL/Ubuntu, but it would be helpful if people can test it in a proper Linux environment and on MacOS. 

- run `yarn test:tz`; all tests should pass. 
- uncomment [this test](https://github.com/HerbCaudill/coronadatascraper/blob/cedf345d1e63c13abd16aeb2dc4cbdf73aa678ee/src/events/crawler/lib/datetime.test.js#L11-L16) in `datetime.test.js` 
- run `yarn test:tz`: now the tests should fail for `Australia/Sydney`. 

### Summary of changes

- **New dependency**: `@js-joda/timezone`. A reduced-size build is imported (only 33KB unminified) that only includes tz data for the current year ± 5 years. The total impact of `js-joda` (core & timezone packages) should be 452KB unminified (207KB minified). 
- New `isoDate` module & tests
- Direct references from the ESP scraper to `js-joda` are replaced with references to `isoDate`.
- **New dev dependency**: `windows-iana` (needed when setting timezones in tests)
- New script `scripts/testTimezone.js` (and associated worker file)

### Why js-joda and not Moment/Luxon/Spacetime/date-fns etc.? 

Moment & Luxon etc. are great for what they do, but they don't offer a date object that has no time associated with it. What sets js-joda apart is that it separates out these concerns:

- [LocalDate](http://js-joda.github.io/js-joda/manual/LocalDate.html) for when you only care about calendar date, not time or time zone 👈 **most of this project's use cases**
- [LocalTime](http://js-joda.github.io/js-joda/manual/LocalTime.html) for when you only care about time of day, not date, and not in any particular time zone
- [LocalDateTime](http://js-joda.github.io/js-joda/manual/LocalDateTime.html) for when you care about time of day + calendar date, but not time zone
- [ZonedDateTime](http://js-joda.github.io/js-joda/manual/ZonedDateTime.html) for when you care about time of day + calendar date in a specific time zone

See the [commented-out test case](https://github.com/HerbCaudill/coronadatascraper/blob/cedf345d1e63c13abd16aeb2dc4cbdf73aa678ee/src/events/crawler/lib/datetime.test.js#L11-L16) mentioned above; This problem is something you still have to deal with in Moment etc., because there's fundamentally no way of getting away from having a time and a time zone associated with a date. You can set that to 0:00:00 and UTC all you want, but then you're always going to have to worry about the date getting converted to someone else's time zone, even if you don't care about time or time zones at all.


src/events/crawler/lib/isoDate.js
```diff
@@ -0,0 +1,54 @@
...
+  if (typeof d === 'string') {
+    const s = truncate(normalize(d));
+    // ISO date
+    if (looksLike.isoDate(s)) return LocalDate.parse(s).toString();
+    // Other formats
+    if (looksLike.YYYYMD(s)) {
+      const ymd = s.split('-'); // e.g. [2020,3,16]
+      return LocalDate.of(...ymd).toString();
+    }
+  }
```
camjc: How to handle fall-through? Should we throw or is undefined fine?

src/events/crawler/lib/datetime.test.js
```diff
@@ -1,6 +1,20 @@
+/* eslint-disable jest/no-commented-out-tests */
```
camjc: I think test.skip won’t complain?

scripts/testTimezones.js
```diff
@@ -0,0 +1,118 @@
+﻿/* eslint-disable import/no-extraneous-dependencies */
```
camjc: Prefer just doing this for a line; then actual mistakes will still be caught in this file

scripts/testTimezones.js
```diff
@@ -0,0 +1,118 @@
...
+// timezone-related heisenbugs. For example, this will fail if the system runtime is east of GMT:
+//
+// ```js
+// test('timezone heisenbug', () => {
+//   const d = new Date('2020-03-16');
+//   expect(getYYYYMMDD(d)).toBe('2020-03-16'); // sometimes we get 2020-03-15
+// });
+// ```
+
+// WARNING: On Windows, this script actually changes the system timezone. If the script doesn't exit
```
camjc: Wow, Windows is weird 😮

scripts/testTimezones.js
```diff
@@ -0,0 +1,118 @@
...
+      process.exit(2);
+    }); // clean up when interrupted with ctrl-C
+  } else {
+    // on linux etc. we can just set this environment variable
+    process.env.TZ = tz;
+  }
+};
+
+const restoreTZ = () => {
+  const tz = originalSystemTimezone;
```
camjc: Why set this const and not just use originalSystemTimezone? Could be clearer.

scripts/testTimezones.js
```diff
@@ -0,0 +1,118 @@
...
+  if (tz && tz !== currentSystemTimezone()) {
+    setTZ(tz);
+  }
+};
+
+const lookupTZ = inputTZ => {
+  if (inputTZ === 'UTC') return inputTZ;
+  const tz = isWindows
+    ? ianaWin.findWindows(inputTZ) //
```
camjc: Stray //

src/events/crawler/lib/isoDate.js
```diff
@@ -0,0 +1,54 @@
...
+const currentJsDate = () => new Date(Date.now()); // allows us to mock current date
+const currentZdt = () => ZonedDateTime.parse(currentJsDate().toISOString());
+const normalize = d => d.replace(/[\\/.]/g, '-');
+const truncate = d => d.slice(0, 10); // truncate ISO datetime to ISO date
+
+// These don't check that string is a valid date, just that it has the right form
+export const looksLike = {
+  isoDate: s => /^\d{4}-\d{2}-\d{2}$/.test(s), // YYYY-DD-MM
+  YYYYMD: s => /^\d{4}-\d{1,2}-\d{1,2}$/.test(s) // YYYY-D-M
+};
```
camjc: There’s no built in checks in js-joda? If there is that may be preferred

src/events/crawler/lib/isoDate.js
hyperknot: I think a console.error with that red cross is a good idea here. Or throwing if the whole app is handling throwing consistently.

src/events/crawler/lib/isoDate.js
```diff
@@ -0,0 +1,54 @@
+﻿/* eslint-disable camelcase */
+
+import { ZonedDateTime, LocalDateTime, LocalDate, ZoneId } from '@js-joda/core';
+import '@js-joda/timezone/dist/js-joda-timezone-10-year-range'; // minimize package size by only importing tz data for current year ±5 yrs
+
+// util functions
+const currentJsDate = () => new Date(Date.now()); // allows us to mock current date
```
hyperknot: I'd use UTC for now and today. What is the point in having different times based on a server's or client computer's locale settings.

I recommend making the codebase 100% UTC, with conversions to and from the location's timezone where needed.

hyperknot: I mean I agree with the joda object not having a timezone. It's just that the "local" time we are working with should always be UTC local.

hyperknot:
Looks nice to me and really clean code!

src/events/crawler/lib/datetime.test.js
HerbCaudill: that offends a different eslint rule, but it's cleaner. I'll change it.

scripts/testTimezones.js
HerbCaudill: 👍

scripts/testTimezones.js
HerbCaudill: that's there to force prettier to break - feels more legible to me?

src/events/crawler/lib/isoDate.js
HerbCaudill: no, you'd just have to try parsing and catch an error.

scripts/testTimezones.js
camjc: Seems like a hack to me but at least a minimal one.

src/events/crawler/lib/isoDate.js
HerbCaudill: yeah, that's probably the right call.

scripts/testTimezones.js
HerbCaudill: 🤷‍♀️🙂

src/events/crawler/lib/isoDate.js
HerbCaudill: I think it should throw - I'll add that.

scripts/testTimezones.js
```diff
@@ -0,0 +1,118 @@
...
+  return tz;
+};
+
+const pause = (t = 500) => new Promise(ok => setTimeout(() => ok(), t));
+
+// Execution
+
+run().then(result => {
+  process.exit(result);
+});
```
qgolsteyn: Could we add a catch and reset the timezone in case we changed it?

scripts/testTimezones.js
```diff
@@ -0,0 +1,118 @@
...
+// ```js
+// test('timezone heisenbug', () => {
+//   const d = new Date('2020-03-16');
+//   expect(getYYYYMMDD(d)).toBe('2020-03-16'); // sometimes we get 2020-03-15
+// });
+// ```
+
+// WARNING: On Windows, this script actually changes the system timezone. If the script doesn't exit
+// normally, you may be left in the wrong timezone. You can set it back by running this from the
+// command line:
```
qgolsteyn: Would it be possible to add a process.on(`uncaughtException`) to reset the timezone in case we changed it?

hyperknot:
Why don't we just skip timezone tests on Windows? Tests are automated on CI anyway, I wouldn't run them on Windows.

qgolsteyn:
Yes, that makes sense. Would make testing locally less brittle.

scripts/testTimezones.js
HerbCaudill: The more likely scenario for ending up with the wrong timezone is that the whole process gets interrupted (e.g. terminal is closed), and there's nothing we can do about that in code. I think @hyperknot is right, we can add a step to run these in CI, and a developer who is making code changes that touch timezones can run `yarn test:tz` as needed, and be aware that it might mess with their computer's time. As long as a normal `yarn test` doesn't run these I think it's fine.

scripts/testTimezones.js
HerbCaudill: see next comment

qgolsteyn:
Could we skip timezone testing on Windows? Our CI runs on MacOS so it should not be a problem. After this we can merge this PR.

HerbCaudill:
I'm on Windows and I need to be able to run these tests sometimes. So my thinking was that timezone testing would only be run manually (and in CI); the only way those tests would run on a developer's machine would be if they specifically did `yarn test:tz` - these tests wouldn't run with a normal `yarn test`. 

Does that make sense? Let me know what you think -

qgolsteyn:
This makes sense! If you can add timezone testing to the CI (by modifying `.github/workflows/test.yml`, I am ready to approve after that.

ryanblock:
Strongly disagree that Windows devs should only be able to work confidently in some, not all, of the codebase. Also noticed a lot of hardcoded *nix (`/`) paths instead of using `path.join` (which I've been trying to clean as I find them). I know it's a bit of a pain to fully support Windows when the app isn't actually running on Windows, but it's the right thing to do.

HerbCaudill:
@ryanblock This way there's no difference between Windows devs and others - everyone has to run `test;tz` independently if they want to run the date tests across multiple timezones.

HerbCaudill:
> This makes sense! If you can add timezone testing to the CI (by modifying .github/workflows/test.yml, I am ready to approve after that.

@qgolsteyn done!

site/lib/color.js
```diff
@@ -9,7 +9,7 @@ export const noCasesColor = '#ffffff';
 export const noPopulationDataColor = 'rgba(0,0,0,0)';
 
 export const outlineColorHighlight = 'rgb(0,0,0)';
-export const outlineColor = 'rgba(0, 0, 0, 0.3)';
+export const outlineColor = 'rgba(0, 0, 0, 0)';
```
qgolsteyn: This should be in a different PR.

HerbCaudill: oops - sorry about that! snuck in from #419

PR #397, opened 2020-03-26
feat: add a QA reporting tool

qgolsteyn:
This PR starts where #291 stopped and creates a QA tool to quickly verify the accuracy of our data set. After timeseries has been run, we use the `timeseries-byLocation.json` output to perform the following tests:

- are data values increasing or remaining stable on a day by day basis?
- are there any missing days (skips of 2 or more days)?
- are data values valid (integer, no NaN, not negative)?
- are the number of tests greater than the number of cases, are the number of deaths fewer than the number of cases?

Results are written to `dist/data-quality.json` in JSON with the following structure: location/test/test message.

Currently, we are performing around 715,000 assertions during the QA process. Jest was not performant for this task. I used the [`zora`](https://github.com/lorenzofox3/zora) library which I found to be the most performant Node testing framework out there. The QA script takes 11s to complete, and currently finds 3,045 errors.

While introducing another testing framework for the project may be an issue, I recommend keeping Jest as the testing framework for the rest of this project, as more people are familiar with it, and the performance requirements are not as stringent.


chunder:
Can you share a the `dist/data-quality.json` that it currently produces?

One concern I have is that 3,045 errors a hopeless number. Are all of these actual errors that need to be resolved?

qgolsteyn:
Here you go: https://gist.github.com/qgolsteyn/4fdf5f5a660f0f2a9d6221db20342137

Most of these errors come from missing dates. All test could be changed to only include the last few days so we only start to apply QA after a certain time.

chunder:
This is great. I was just sitting down to start working on something like this.

Good idea to run it only on the last few days to pick up regressions.

I might also also relax the monotonic constraint. It's pretty common for agencies to revise estimates +/- 2 or something if they made an error or whatnot.

chunder:
When run for the last 5 days, will all the missing data windows that started way back around 3/11 stop getting reported? Because that's most of them. I think caused by JHU dropping detailed reporting? Making a new scraper for one of these basically triggers a huge missing-data window.

On the monotonic constraint.. I feel like we might want to highlight:

  - A value drops to 0. That's bad, even if it went from just 1->0.
  - A value drops by more than 2 or 3 -- that signals more than just a minor correction from an agency.

And the `deaths <= cases <= tested` test... can it report the values for those items, not just "should be truthy"?

Again, awesome work. Don't take my feedback a sign of anything but full support for your work.

qgolsteyn:
Changes:
- Looking at missing windows from 2020-03-20 onward. Ignoring missing windows that started before 2020-03-20.
- Added a test to detect drops to 0. Allow for a decrease of up to 3 per day for the monotonic constraint.
- deaths <= cases <= tested now gives a more useful message

qgolsteyn:
Updated gist: https://gist.github.com/qgolsteyn/4fdf5f5a660f0f2a9d6221db20342137

PR #400, opened 2020-03-26
Add Russian data crawler

freeatnet:
Hey everyone! This PR adds a crawler for COVID19 spread data coming out of Russia. There are a couple things that may need addressing.

The first is that I see a rule against non-government/health organizations sources. Yandex is not a government organization (it’s kind of a Google of Russia); however, their map is actually linked from the website of the agency tasked with COVID19 spread tracking, https://www.rospotrebnadzor.ru/about/info/news_time/news_details.php?ELEMENT_ID=13566 (the boxy button on the right). Does that make it an acceptable source?

The second is that their JSON feed requires a CSRF token, which is paired with a cookie value. needle is perfectly capable of handling cookies; however, the `fetch.json` convenience method does not expose that functionality. Is it acceptable to use needle directly or do I need to dig in and add cookie support to `fetch.*` methods?


src/events/crawler/scrapers/RUS/index.js
```diff
@@ -0,0 +1,39 @@
+import needle from 'needle';
+
+const scraper = {
+  country: 'RUS',
+  aggregate: 'state',
+  url: 'https://yandex.ru/maps/api/covid?csrfToken=',
```
lazd: Can you include a `curators: [{name: 'Yandex'}]` etc bit to cite Yandex as the curator?

Also, if Yandex cites its sources, can you include `sources: [{name: 'etc', url: 'etc'}]`?

freeatnet: How's this?

freeatnet:
Updated with population counts.

src/events/crawler/scrapers/RUS/index.js
```diff
@@ -0,0 +1,65 @@
...
+      email: 'arseniy+coronadatascraper@freeatnet.com'
+    }
+  ],
+  async scraper() {
+    const csrfRequestResponse = await needle('get', this.url, {}, { parse_response: true });
+    const csrfCookies = csrfRequestResponse.cookies;
+    const { csrfToken } = csrfRequestResponse.body;
+    // assert typeof csrfToken !== 'undefined'
+
+    const { body } = await needle('get', `${this.url}${csrfToken}`, { cookies: csrfCookies, parse_response: true });
```
qgolsteyn: We need the scraper to use `fetch.js` for caching to work. If you look at `fetch()` in `fetch.js`, it accepts an options object. I would suggest passing cookies in that options object, and updating this [line](https://github.com/lazd/coronadatascraper/blob/f86a1193e5564af634a2dbe7637be2c71d621185/src/events/crawler/lib/get.js#L64) to add cookies to the needle request.

qgolsteyn:
Updated to use `fetch`. I can't have it completely use `fetch` as we need to retrieve a cookie from a website.

src/events/crawler/lib/get.js
```diff
@@ -61,7 +62,7 @@ export const get = async (url, type, date = process.env.SCRAPE_DATE || datetime.
       }
 
       let errorMsg = '';
-      const response = await needle('get', url).catch(err => {
+      const response = await needle('get', url, { cookies }).catch(err => {
```
jzohrab: I'm not a real js developer, but the hash feels like it needs a key, something like

```
opts = { cookies: cookies }
const response = await needle('get', url, opts).catch(err => {
```

because `needle(...)` actually gets the cookies by a lookup:

```
  if (options.cookies)
    config.headers['cookie'] = cookies.write(options.cookies);
```

In addition this change will cause a null/undefined cookie to be written for every call to `get` that doesn't include the cookie.  I don't know if that's problematic or not, but it feels wrong.

freeatnet:
Thanks @qgolsteyn!

ngolosov:
@freeatnet 
Could we add data scraping for the Nenets Autonomous Okrug (ISO code RU-NEN) and Chukotka Autonomous Okrug(RU-CHU)? Yandex reports the counts for these districts, but I can't find them in the resulting files. 
Thank you

PR #411, opened 2020-03-27
Added retry logic for headless fetches

chunder:
I noticed that WA, USA was flakey. The server or headless seems to fail <10% of the time. So I added retry logic. I couldn't get any failures to happen to prove it out though. So can someone review this code carefully?

### DIFF
--- a/src/shared/lib/fetch/get.js
+++ b/src/shared/lib/fetch/get.js
@@ -57,7 +57,7 @@ export const get = async (url, type, date = process.env.SCRAPE_DATE || datetime.
       tries++;
       if (tries > 1) {
         // sleep a moment before retrying
-        console.log(`  ⚠️  retrying (${tries})`);
+        console.log(`  ⚠️  Retrying (${tries})...`);
         await new Promise(r => setTimeout(r, 2000));
       }
 
@@ -97,6 +97,10 @@ export const get = async (url, type, date = process.env.SCRAPE_DATE || datetime.
       console.log(`  ❌ Got error ${response.statusCode} trying to fetch ${url}`);
       return null;
     }
+
+    console.log(`  ❌ Failed to fetch ${url} after ${tries} tries`);
+    return null;
   }
+
   return cachedBody;
 };
--- a/src/shared/lib/fetch/index.js
+++ b/src/shared/lib/fetch/index.js
@@ -190,7 +190,7 @@ const fetchHeadless = async url => {
     }
   }
 
-  console.log(`  ❌ Too many trying to fetch ${url}`);
+  console.log(`  ❌ Failed to fetch ${url} after ${tries} tries`);
   return null;
 };
 
--- a/src/shared/lib/fetch/index.js
+++ b/src/shared/lib/fetch/index.js
@@ -153,7 +153,7 @@ const fetchHeadless = async url => {
     tries++;
     if (tries > 1) {
       // sleep a moment before retrying
-      console.log(`  ⚠️  retrying (${tries})`);
+      console.log(`  ⚠️  Retrying (${tries})...`);
       await new Promise(r => setTimeout(r, 2000));
     }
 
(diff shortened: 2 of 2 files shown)


src/events/crawler/lib/fetch.js
```diff
@@ -147,30 +147,49 @@ const fetchHeadless = async url => {
...
+        await page.waitFor(RESPONSE_TIMEOUT);
+        const html = await page.content();
+        browser.close();
+        return html;
+      }
+      if (response._status < 500) {
+        console.log(`  ❌ Got error ${response._status} trying to fetch ${url}`);
+        browser.close();
+        return null;
+      }
```
jzohrab: Will the server ever have an internal server error (code >= 500)?  And if so, should it retry?  E.g., see the checks in https://github.com/lazd/coronadatascraper/blob/master/src/events/crawler/lib/get.js#L80.

src/events/crawler/lib/fetch.js
```diff
@@ -147,30 +147,49 @@ const fetchHeadless = async url => {
...
-  } catch (err) {
-    browser.close();
 
-    if (err.name === 'TimeoutError') {
-      console.log(`  ❌ Timed out trying to fetch ${url}`);
-      return null;
+      if (err.name === 'TimeoutError') {
+        console.log(`  ❌ Timed out trying to fetch ${url}`);
+      } else {
+        console.log(`  ❌ Caught error trying to fetch ${url}`);
```
jzohrab: Consider actually writing error details here.

src/events/crawler/lib/fetch.js
```diff
@@ -147,30 +147,49 @@ const fetchHeadless = async url => {
...
+      }
     }
-    throw err;
   }
+
+  return null;
```
jzohrab: Consider logging something to the effect of "exceeded max retries" here.

jzohrab:
Hi @chunder, I don't know that tech well, so my notes might not be the "careful code review" you're looking for.  Cheers! z

src/events/crawler/lib/fetch.js
jzohrab: (I'm assuming that a >=500 error wouldn't actually result in an error being thrown)

chunder:
@jzohrab I added better comments and logging per your suggestions.

@lazd @qgolsteyn Either you able to eyeball the logic here. I think it's sane, but it's hard to test.

src/events/crawler/lib/fetch.js
lazd: Hmm, I think retrying on an internal server error is likely to just result in another failure (trying 30 minutes later, maybe not), but it's possible we could get it right as it's fixed...

src/events/crawler/lib/fetch.js
```diff
@@ -147,30 +147,50 @@ const fetchHeadless = async url => {
...
-      waitUntil: 'networkidle2'
-    });
-
-    if (response._status < 400) {
-      await page.waitFor(RESPONSE_TIMEOUT);
-      const html = await page.content();
-      browser.close();
-      return html;
+  let tries = 0;
+  while (tries < 5) {
```
lazd:
```suggestion
  let tries = 0;
  let maxTries = 5;
  while (tries < maxTries) {
```

src/events/crawler/lib/fetch.js
chunder: Servers return 503 or other 5xx all the time when they are simply overloaded. Sleeping a bit and retrying is a perfectly great thing to do.

src/events/crawler/lib/fetch.js
```diff
@@ -147,30 +147,50 @@ const fetchHeadless = async url => {
...
+      }
+    } catch (err) {
+      // Caught something, allow retry
+      browser.close();
+      console.log(`  ❌ Caught ${err.name} error trying to fetch ${url}`);
     }
-    throw err;
   }
+
+  console.log(`  ❌ Too many trying to fetch ${url}`);
```
lazd:
```suggestion
  console.log(`  ❌ Failed to fetch ${url} after ${tries} tries`);
```

lazd:
@chunder it does indeed look sane. I made this and the other retry logic the same and added better logging, can you look again at it and check I didn't screw anything up?

PR #419, opened 2020-03-27
viz tweak: transparent outlines on choropleth

HerbCaudill:
I find the gray borders on the choropleth to be really noisy. Especially when zoomed out, areas with a high density of admin boundaries (e.g. county borders in the eastern US) almost acquire an unintended additional map shade. 

Before: 

image

After:

77743353-3ccfd180-7018-11ea-805d-59a0cddf4275

### DIFF
--- a/site/lib/color.js
+++ b/site/lib/color.js
@@ -9,7 +9,7 @@ export const noCasesColor = '#ffffff';
 export const noPopulationDataColor = 'rgba(0,0,0,0)';
 
 export const outlineColorHighlight = 'rgb(0,0,0)';
-export const outlineColor = 'rgba(0, 0, 0, 0.3)';
+export const outlineColor = 'rgba(0, 0, 0, 0)';
 
 const choroplethColors = {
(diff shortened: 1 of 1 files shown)


lazd:
Yeah it's noisy, but it's hard to tell areas apart without it. @NateBaldwinDesign @alangwilson?

NateBaldwinDesign:
Oh that is a good point. We don’t want to imply coloration where there is none. However at a zoomed in level those borders are important— is there a way we could have the borders disappear under certain conditions (ie zoom level)?

lazd:
@NateBaldwinDesign I bet we can make them go away at a given zoom level, we should investigate that!

camjc:
Can/should we close this now we have country/state/county toggles?

hyperknot:
I think the borders are quite nice! Also, a lot has changed since 23 days, we have the toggles now as @camjc said. I recommend closing this.

jzohrab:
Agreed with @hyperknot and @camjc, closing this.  Thanks for the PR though, @HerbCaudill .

Issue #374, opened 2020-03-26
geography.addEmptyRegions can easily mistakenly infill all counties with 0 cases in a state


chunder:
So I just realized that `geography.addEmptyRegions` is a bit dangerous.

If a scraper gets confused by a change in the source page, it will sometimes fail to find any counties. That’s cool, returning no results is a safe/conservative thing to do.

But then most county level state scrapers call `addEmptyRegions`. Which sees no data. And backfills 0 for every county.

I wonder if `addEmptyRegions` should throw an error if it gets called with *no* actual counties.

lazd:
This is true... I agree, we should throw if no data is present!

chunder:
PR created: https://github.com/lazd/coronadatascraper/pull/421

chunder:
Now we won't be outputting `0` cases if a scraper breaks. Will the map properly reflect the prior day's data if the time series has a missing entry for the final day?

lazd:
Fixed in #374

PR #421, opened 2020-03-27
Added throw if addEmptyRegions called without any actual data

chunder:
Fix for issue described here: https://github.com/lazd/coronadatascraper/issues/374


lazd:
@chunder next time write `Fixes #374` and Github will close the issue automatically on merge :D

PR #425, opened 2020-03-27
fix: IN has no test data, MD reports tested as recovered

qgolsteyn:


src/events/crawler/scrapers/USA/MD/index.js
```diff
@@ -50,7 +50,7 @@ const scraper = {
           county: countyName,
           cases: parse.number(county.COVID19Cases),
           deaths: parse.number(county.COVID19Deaths),
-          tested: parse.number(county.COVID19Recovered)
```
lazd: Oops

PR #428, opened 2020-03-27
Subregion toggle

jflaboe:
UI addition to toggle county-level and state-level data on and off for GeoJSON.

Reveals missing state/country data that might be able to be pieced together using aggregation.


lazd:
@jflaboe can you send this against the `mapupdates` branch? I was working on adding a timeseries slider and touched some of the same places.

jflaboe:
Moving to new pull request to compare against mapupdates branch
https://github.com/lazd/coronadatascraper/pull/432

PR #432, opened 2020-03-27
Subregion-toggle

jflaboe:


site/map.js
```diff
@@ -38,35 +38,52 @@ function initData() {
...
     type: 'FeatureCollection',
     features: data.features.features.filter(function(feature) {
       const location = data.locations[feature.properties.locationId];
-      if (location && !location.county && location.aggregate === 'county') {
-        return false;
+      if (showSubregion) {
+        if (location && !location.county && location.aggregate === 'county') {
+          return false;
+        }
       }
```
camjc:
```suggestion
      if (showSubregion && location && !location.county && location.aggregate === 'county') {
        return false;
      }
```

site/map.js
```diff
@@ -38,35 +38,52 @@ function initData() {
...
   const countryFeatures = {
     type: 'FeatureCollection',
     features: data.features.features.filter(function(feature) {
       const location = data.locations[feature.properties.locationId];
-      if (location && location.country && !location.state && location.aggregate === 'state') {
-        return false;
+      if (showSubregion) {
+        if (location && location.country && !location.state && location.aggregate === 'state') {
+          return false;
+        }
```
camjc: Feels repetitious of the generateStateFeatures, should/can we abstract?

camjc:
Looking cool 👍

praging:
See https://github.com/covidatlas/coronadatascraper/issues/616 for a possible issue.

jzohrab:
@lazd , should this get merged into the mapupdates branch (last commits of this were from about 2 weeks ago : https://github.com/covidatlas/coronadatascraper/commits/mapupdates )?

This PR is going into that feature branch, and not into master.  There are some questions on it, but it looks like it could keep that branch moving forward.  Thoughts?

lazd:
@jzohrab not sure, but we should look at this and #367 and decide if we want those features on our map, or if we should move that development to the COVIDAtlas frontend work?

lazd:
Refactored this a bit and merges this into mapupdates, thanks!

PR #435, opened 2020-03-27
Fixed WY, USA scraper

chunder:


lazd:
Excellent, looks to be backwards compatible too.

PR #437, opened 2020-03-27
[patch] fix AUS SA, they have a table now hooray

camjc:
image

### DIFF
--- a/src/shared/scrapers/AUS/SA/index.js
+++ b/src/shared/scrapers/AUS/SA/index.js
@@ -42,7 +42,7 @@ const scraper = {
       };
     },
     '2020-3-27': async function() {
-      scraper.type = 'table';
+      this.type = 'table';
       const $ = await fetch.page(this.url);
       const $table = $('table:first-of-type');
       const $trs = $table.find('tbody > tr:not(:first-child)');
--- a/src/shared/scrapers/AUS/SA/index.js
+++ b/src/shared/scrapers/AUS/SA/index.js
@@ -2,6 +2,20 @@ import * as parse from '../../../lib/parse.js';
 import * as fetch from '../../../lib/fetch/index.js';
 import maintainers from '../../../lib/maintainers.js';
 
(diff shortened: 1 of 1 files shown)


src/shared/scrapers/AUS/SA/index.js
```diff
@@ -2,6 +2,20 @@ import * as parse from '../../../lib/parse.js';
 import * as fetch from '../../../lib/fetch/index.js';
 import maintainers from '../../../lib/maintainers.js';
 
+const getKey = rowLabel => {
+  const lowerLabel = rowLabel.toLowerCase();
+  if (lowerLabel.includes('confirmed cases')) {
+    return 'cases';
+  }
+  if (lowerLabel.includes('icu')) {
+    return 'hospitalized';
```
camjc: random guess, if we have a standard name for this please let me know.

src/shared/scrapers/AUS/SA/index.js
```diff
@@ -17,14 +31,29 @@ const scraper = {
...
+      const $ = await fetch.page(this.url);
+      const paragraph = $('.middle-column p:first-of-type').text();
+      const { casesString } = paragraph.match(/been (?<casesString>\d+) confirmed cases/).groups;
+      return {
+        state: scraper.state,
+        cases: parse.number(casesString)
+      };
+    },
+    '2020-3-27': async function() {
+      scraper.type = 'table';
```
camjc: Is there a better way to change type per date of scraper?

src/shared/scrapers/AUS/SA/index.js
lazd: We don't, but that's a good start!

src/shared/scrapers/AUS/SA/index.js
lazd: Nah, just `this.type` (`scraper.type` works ofc, too)

PR #439, opened 2020-03-27
[patch] Simplify AUD QLD since HHS areas arent mappable yet

camjc:
image
We don't have a way to map the HHS regions yet, so lets just simplify to fix the map for now.


lazd:
Sounds good to me; if we want to go back and break it down later, we'll have cache to do so.

PR #440, opened 2020-03-27
feat: request hospitalized data

camjc:


lazd:
So is `hospitalized` cumulative or not? If it's cumulative, then we need a `discharged` field... If it's not cumulative, then it's the only field in the dataset that isn't...

lazd:
Once we figure that out, can you also update https://github.com/lazd/coronadatascraper/blob/master/docs/sources.md and include `hospitalized`?

PR #446, opened 2020-03-28
[patch] fix lint in master

camjc:
image


jzohrab:
LGTM as a fix, though the code likely should be able to handle this case (e.g. for fresh builds).

PR #449, opened 2020-03-28
Fixed Clark County, NV, USA

chunder:


ryanblock:
Tested locally, looks good!

PR #452, opened 2020-03-28
Fixed PA, USA. Looks like it has been silently broken for awhile, col…

chunder:
…lecting stats from the wrong table


ryanblock:
Tested locally, looks good!

PR #453, opened 2020-03-28
Sonoma,CA has been broken and returning 0 for awhile

chunder:


lazd:
Let's make a `DeprecatedError` class in `src/shared/lib/errors.js` and throw an instance of that:

```
export class DeprecatedError extends Error {
  constructor(message) {
    super(message);
    this.name = "DeprecatedError";
  }
}
```

Then, in the scraper reports, we can do `if (err instanceof DeprecatedError) { doNothing(); }`

@qgolsteyn @camjc ?

camjc:
In a PR I just did I returned an empty object. Good to have a standard way of doing this.

camjc:
Yup, turns out an empty object still reports. Dang. Let’s do your approach.

lazd:
We'll go ahead and close this one, thanks!

PR #462, opened 2020-03-28
fix: wisconsin cases/deaths after 3/25

### LINKED ISSUE #426: Wisconsin cases on 3/26 are incorrect
Pulled the latest data and it seems that the Wisconsin cases are actually showing the death values from  https://www.dhs.wisconsin.gov/covid-19/data.htm.   Looks like the data from 3/25 and prior is still okay

cmcjacob:
Fixes #426 

timeseries accurate with new cache (tested from 3/12 - 3/28)


lazd:
Thank you!

Issue #431, opened 2020-03-27
Pretty names for sources (`url` field)


Akuukis:
## Problem
It's great to have `url` field to see which data comes from which places!

It would be even better to have a "pretty name" for those urls. For example, for "https://github.com/CSSEGISandData/COVID-19" says "Johns Hopkins University" or so.

## Proposal
A non-breaking suggestion would be to provide a map where urls maps to pretty names.

A breaking suggestion would be to change `url` field to `source` field with interface `{title: string, url: string}`. 

## Rationale

I want to present the source to users but they care more about the owner of source than the link to the source.

lazd:
We already have a `sources` field (as well as `curators` for curated date sources), but many scrapers don't have this field populated!

@Akuukis, can you send a PR adding pretty names to every scraper? We will be forever indebted to you.

Akuukis:
@lazd I can look into that but my earliest ETA would be end of Monday. If anyone can do it sooner then please do!

lazd:
Sounds great!

jzohrab:
Hi @Akuukis , this is a nice idea.  Did you make any progress on this?  If not, what would help you make progress on it?  Cheers, regards! jz

Akuukis:
@jzohrab Sorry, no, I didn't. In the end I figured it is not that big a deal for my dashboard, so I skipped this. Sorry for not coming back sooner!

jzohrab:
Np, thanks for replying @Akuukis.  I'll keep this issue open as it's still valid.

What's your dashboard?  I keep a list of consumers for interests' sake.

Akuukis:
@jzohrab That's [Pandemic Estimator](https://pandemic-estimator.net)

jzohrab:
Thanks very much, that's an interesting presentation!  Are you still using us for a source?  I see on the graph footer a link to other sources, eg https://health-infobase.canada.ca/src/data/covidLive/covid19.csv for Canada.

Akuukis:
@jzohrab Thanks! Yes, it's mentioned on "Methodology" page

jzohrab:
Closing, we're addressing this in Li.  cheers! z

PR #463, opened 2020-03-28
Added some of the missing sources.

jacobmcgowan:
Made some progress on #431 but does not resolve it yet as there are still many more scrappers that need sources filled out.


lazd:
Fantastic! Thank you!

PR #467, opened 2020-03-28
Fix fresno county total cases

livejake:
Currently the Fresno county scraper is concatenating Total cases and Travel related cases to report 3116 cases instead of 31. This fixes the issue.


lazd:
Nice, that's clever! Remove the children, go back up to the previous filtering state, and return text. Should be backwards and forwards compat too.

PR #476, opened 2020-03-29
Add county level scraper & GeoJSON for Brazil

thiagoghisi:
A lot of kudos to https://brasil.io/datasets team and @turicas (https://github.com/turicas/covid19-br) that have put countless hours and organized many volunteers in Brazil to make it possible to have a centralized CSV for the entire country (all cities). Pretty much a similar project as this one, but only for Brazil.

Unfortunately in Brazil, the government website with the data is down for the past 2 days…

That is the main reason I had to switch the Brazilian Scraper from `http://plataforma.saude.gov.br/novocoronavirus/resources/scripts/database.js` to `https://brasil.io/dataset/covid19/caso?place_type=city&format=csv`

This is my first PR here so critical feedback would be much appreciated.

image


turicas:
If you need only the latest confirmed/deaths for each city, you must pass `is_last=True` in the querystring.

thiagoghisi:
> If you need only the latest confirmed/deaths for each city, you must pass `is_last=True` in the querystring.

Hey @turicas, thanks for letting me know! But, I'm actually adding time-series support here, so I need past data for the scrapper as well so we can see the history of growth and also simulate old dates.

thiagoghisi:
Fixing the lint problems now... Having some problems to run lint locally.

lazd:
This is an extensive piece of work! It'll take some time to review, but we'll get it in!

thiagoghisi:
@lazd Alright Larry, this PR is all yours now. 

Let me know if you have any idea how to compress or minify the GeoJSON data file for all Brazilian counties (it is pretty massive... almost 60MB. :-().

Also, there are a few opportunities for refactoring on `find-features` and `find-population`, but for now, I decided to follow the existing patterns duplicating pretty much the same code that was being used for US & UK counties mapping. Probably ok for now, but something to think about as we add more counties date for more countries.

Thanks a lot!
Thiago.

turicas:
@thiagoghisi I've simplified all Brazilian GeoJSONs made available by IBGE and they are lighter now. I'm going to publish them in Brasil.IO soon.

thiagoghisi:
That is awesome tip @turicas! Thanks a lot for sharing, I will try to make this change either tonight or tomorrow.

lazd:
@thiagoghisi shall I wait to merge until the change is in?

thiagoghisi:
@lazd I'm totally fine merging it as it is and then I will open a separate PR with both the GeoJson optimizations & the state level data aggregation following a similar model that was done for RI State.

lazd:
@thiagoghisi ok, let's do it!

lazd:
Dang, getting an error: maybe the API changed?

```
✅ All scrapers are valid!
  🐢  Cache miss for http://plataforma.saude.gov.br/novocoronavirus/resources/scripts/database.js?v=2020-03-30 at cache/b0ca850ff9e8dc197d76ddf938eb01e0.html
  🚦  Loading data for http://plataforma.saude.gov.br/novocoronavirus/resources/scripts/database.js?v=2020-03-30 from server
  ❌ Error processing BRA:  TypeError: Cannot read property 'text' of null
    at Object.scraper (/Volumes/lazdRAID/Users/lazdHome/repos/coronavirus-data-sources/scraper/src/shared/scrapers/BRA/index.js:1108:38)
    at async runScrapers (/Volumes/lazdRAID/Users/lazdHome/repos/coronavirus-data-sources/scraper/src/events/crawler/scrape-data/run-scraper.js:103:38)
    at async generate (/Volumes/lazdRAID/Users/lazdHome/repos/coronavirus-data-sources/scraper/src/shared/index.js:32:18)
```

lazd:
I would recommend making it a separate scraper file `BRA/county-level.js` so we can have both the official government source from #548 as well as this.

Once we get that error fixed and get the smaller GeoJSON in, then rebase and move to a new scraper file, we can merge this!

thiagoghisi:
Sounds good @lazd!

The error you were getting is related to the version of the Brazilian scraper that was still on master, not my version on this branch.

hyperknot:
Hey, some things related to the GeoJSON:

1. How did you simplify the GeoJSON?
2. Can you name it .geojson, not .json? GitHub for example has super nice built-in viewer for .geojson extensions.
3. Whatever we do, let's not check in 60 MB to GitHub. It'll make the .git repo even bigger for everyone, even if we delete it afterwards.

For the county codes, do I understand it right that there is a system in Brazil similar to FIPS codes in the US?

BTW, I'm thinking it might be better to merge the population and GeoJSON files to the [country-levels](https://github.com/hyperknot/country-levels) repo I'm developing. That'd allow us to have multiple quality levels.

thiagoghisi:
Hey Zsolt, great questions! Amazing review. Let me go one by one:

> 1. How did you simplify the GeoJSON?

Multiple rounds of `JQ` transformations on the terminal. A few helpful ones that I used over the weekend after digging in:
`$ jq -c '.features[] |= del(.properties.id, .properties.populacao, .properties.pib, .properties.estado_id, .properties.codigo_ibg)' bra-counties.json > bra-counties-stripped.json`
`$ jq -c '.features | map( {geometry: .geometry, properties: (.properties + { "name": (.properties.nome + ", " + .properties.uf) }), type: .type} )’ bra-counties.json > bra-counties-stripped.json`

> 2. Can you name it .geojson, not .json? GitHub for example has super nice built-in viewer for .geojson extensions.

💯! I will do that in a separate PR also renaming the US, Italy and other geojson files on this repo committed as .json.

> 3. Whatever we do, let's not check in 60 MB to GitHub. It'll make the .git repo even bigger for everyone, even if we delete it afterwards.

Totally agree. I will fix this on this PR and do a force push on my branch so no one has this massive file on their git tree.

> For the county codes, do I understand it right that there is a system in Brazil similar to FIPS codes in the US?

No, Brazil has CEP that is similar to US Zip Codes.

> BTW, I'm thinking it might be better to merge the population and GeoJSON files to the [country-levels](https://github.com/hyperknot/country-levels) repo I'm developing. That'd allow us to have multiple quality levels.

Definitely not against it. I'm totally fine moving it there along with the remaining `.geojson` files to your repo.

Thanks,
Thiago.

hyperknot:
Hey Thiago,

> Multiple rounds of JQ transformations on the terminal.

I see so you didn't actually simplify the geometry, but you've cleaned up the properties a bit right? The really complicated things with GeoJSON is to simplify the geometry, which is really tricky actually, and involves converting to TopoJSON and back. Here is how I do it on full planet OpenStreetMap:
https://github.com/hyperknot/country-levels/blob/master/scripts/wam_topo_simplify.sh

> Brazil has CEP that is similar to US Zip Codes.

Perfect. We are just moving to an id system, so US counties will be like `fips:06085` for example. So for Brazil, we can do a similar naming like `br_cep:23167` right?

> I'm totally fine moving it there along with the remaining .geojson files to your repo.

Great. So after the migration to the id system + country-levels, there won't be any single GeoJSON file in the main repository. It'd be really nice if we could move yours over as well.

Do you want to have a go at adding this br_cep system to https://github.com/hyperknot/country-levels? 

Basically you can copy the fips parts and make them br_cep. But it can be much simpler as you can use your GitHub sources, unless you want to directly download from some official agency, like fips codes to do from US Census.

thiagoghisi:
> Hey Thiago,
> 
> > Multiple rounds of JQ transformations on the terminal.
> 
> I see so you didn't actually simplify the geometry, but you've cleaned up the properties a bit right? The really complicated things with GeoJSON is to simplify the geometry, which is really tricky actually, and involves converting to TopoJSON and back. Here is how I do it on full planet OpenStreetMap:
> https://github.com/hyperknot/country-levels/blob/master/scripts/wam_topo_simplify.sh

Awesome! I was actually planning to use https://github.com/jczaplew/geojson-precision to reduce the precision and consequently the size of my GeoJSON file. Is this a similar thing that your script is doing?

> > Brazil has CEP that is similar to US Zip Codes.
> 
> Perfect. We are just moving to an id system, so US counties will be like `fips:06085` for example. So for Brazil, we can do a similar naming like `br_cep:23167` right?

Not really, CEP is as granular as the US Zip code. This means that 1 county or 1 city will have sometimes hundreds of different CEPs that I'm not sure if there are even GeoJSONs available out there at the CEP level. :-(

Makes sense? 

US FIPS code is different because it is a 1:1 mapping 1 county = 1 FIP. Brazilian CEP is like the US Zip Codes. many to one (*:1) mapping, 1 county has many CEPs or many Zip Codes. Also, besides Singapore, I don't know any country or state that is reporting COVID-19 cases at the Zip-code level...

> 
> > I'm totally fine moving it there along with the remaining .geojson files to your repo.
> 
> Great. So after the migration to the id system + country-levels, there won't be any single GeoJSON file in the main repository. It'd be really nice if we could move yours over as well.

> Do you want to have a go at adding this br_cep system to https://github.com/hyperknot/country-levels?

Sure thing! Working on it now.

> Basically you can copy the fips parts and make them br_cep. But it can be much simpler as you can use your GitHub sources, unless you want to directly download from some official agency, like fips codes to do from US Census.

As I mentioned above, this would not really be possible for Brazil... It would be a massive work and I don't think there are CEP level GeoJSON available for Brazil at the moment, unfortunately.

thiagoghisi:
> I would recommend making it a separate scraper file `BRA/county-level.js` so we can have both the official government source from #548 as well as this.
> 
> Once we get that error fixed and get the smaller GeoJSON in, then rebase and move to a new scraper file, we can merge this!

Alright, Larry, both changes are in! Renamed my scraper, squashed all commit and rebased it.

Also, FYI GeoJSON went from 60mb to 15mb by reducing the geojson-precision to 6 using https://github.com/perrygeo/geojson-precision

src/shared/scrapers/BRA/county-level.js
```diff
@@ -0,0 +1,5621 @@
...
+
+const scraper = {
+  country: 'BRA',
+  type: 'csv',
+  priority: 1,
+  url: 'https://brasil.io/dataset/covid19/caso?place_type=city&format=csv',
+  timeseries: true,
+  aggregate: 'county',
+
+  _counties: [
```
qgolsteyn, line 18: To keep this scraper small (and help with review!) would it be possible to move this county list to a JSON file and import it to here? You can just add this file to the directory where the scraper is.

src/shared/scrapers/BRA/county-level.js
```diff
@@ -0,0 +1,5621 @@
...
+      latestDate = datetime.getYYYYMMDD(scrapeDate);
+    }
+
+    let counties = data
+      .filter(row => {
+        return row.date === latestDate;
+      })
+      .map(row => {
+        return {
+          cases: parse.number(row.confirmed) || 0,
```
qgolsteyn, line 5608: If the number is undefined, don't return 0, we should be leaving the field undefined.

src/shared/scrapers/BRA/county-level.js
```diff
@@ -0,0 +1,5621 @@
...
+          deaths: parse.number(row.deaths) || 0,
```
qgolsteyn, line 5609: If the number is undefined, don't return 0, we should be leaving the field undefined.

src/shared/scrapers/BRA/county-level.js
thiagoghisi: Sure thing! Great idea!

hyperknot:
> Awesome! I was actually planning to use https://github.com/jczaplew/geojson-precision to reduce the precision and consequently the size of my GeoJSON file. Is this a similar thing that your script is doing?

Yes, it's also running geojson-precison in the last step. But before that, it needs to simplify the geometry, similar to how you can see in this demo:
https://mourner.github.io/simplify-js/

>  I think it will be better to move all of them together to your repo if that is the plan to have GeoJSON files outside this repo.

It'll be a gradual change, in the following days I believe.

Now about the actual ID system. Basically it doesn't matter for us what is the underlying structure, there just needs to be some kind of consistent way we can identify features with a short code. What do you think would be the best one? 

We could go with a slugifiied name, for example:
`Lagoa da Confusão, TO` -> `to-lagoa_da_confusao`

Does that look reasonable for you?

src/shared/scrapers/BRA/county-level.js
thiagoghisi: Done! Thank You for the suggestion!

thiagoghisi:
> > Awesome! I was actually planning to use https://github.com/jczaplew/geojson-precision to reduce the precision and consequently the size of my GeoJSON file. Is this a similar thing that your script is doing?
> 
> Yes, it's also running geojson-precison in the last step. But before that, it needs to simplify the geometry, similar to how you can see in this demo:
> https://mourner.github.io/simplify-js/

That is perfect! I totally missed that when looking at your script. Do you still think this would be a good idea to run your script on my GeoJSON file here now that it has around ~16MB? How much more do you think it would reduce in size?

> > I think it will be better to move all of them together to your repo if that is the plan to have GeoJSON files outside this repo.
> 
> It'll be a gradual change, in the following days I believe.

Sounds good! 

> Now about the actual ID system. Basically it doesn't matter for us what is the underlying structure, there just needs to be some kind of consistent way we can identify features with a short code. What do you think would be the best one?
> 
> We could go with a slugified name, for example:
> `Lagoa da Confusão, TO` -> `to-lagoa_da_confusao`
> 
> Does that look reasonable for you?

That is totally reasonable! Do you have any script to do that?

src/shared/scrapers/BRA/county-level.js
thiagoghisi: Done!

hyperknot:
@qgolsteyn we were in the middle of a work on this with @thiagoghisi, wanted to make an id system and wanted to avoid merging the big files to master branch.

hyperknot:
What should we do now? Revert and do the fix tomorrow?

qgolsteyn:
Sure, I can revert

lazd:
We need to simplify the GeoJSON here, `features.json` is getting massive!

hyperknot:
@lazd we are just about to do that with @thiagoghisi

PR #481, opened 2020-03-29
All tests ported to Tape

ryanblock:
Tape is a very simple, widely used, battle-hardened test framework that's part of the language-agnostic TAP (Test Anything Protocol) family.

Unlike many modern test frameworks, like Jest, Tape does not rewire builtin Node.js functionality (such as `require` and `console.log`), or unpredictably mutate globals.

Tape just makes it very easy to make assertions with minimal API surface, which is all tests are supposed to do.

In this PR I ported all of our existing tests from Jest to Tape. Below you will find a detailed comparison.

Notes:

- You can test this branch by running `npm run test:jest` (Jest) `npm run test:tape` (Tape) (or both, back to back, with `npm run test`)
- You may notice the test counts are different; again, these are 1:1 test ports
  - I changed as little as possible to port things over; Jest also appears counts more things (like test statement invocations)

---

## Time

Average Jest run (recent MBP):
```
real	0m10.539s
user	0m18.423s
sys	0m5.599s
```

Average Tape run (recent MBP):
```
real	0m1.531s
user	0m1.735s
sys	0m0.272s
```

Difference: on this test suite Jest is **~7x slower and requires >10x the compute time** (without doing any optimization of Tape runs) to accomplish the same tests

## Test surface area

Jest
- 13 files
- 971 lines of code

Tape
- 8 files
- 522 lines of code

Difference: Jest impl has **~62% more files and ~54% more lines of code** to accomplish the same tests

## Dependency weight

Jest + related dependencies I found:

- `@babel/core`
- `@babel/preset-env`
- `babel-eslint`
- `babel-jest`
- `eslint-plugin-jest`
- `jest`
- `jest-each`
- 387 total dependencies + subdependencies
- Total weight of dependencies: 74.8MB

Tape dependencies:

- `tape`
- `tap-spec` (optional reporter)
- 71 total dependencies + subdependencies
- Total weight of dependencies: 12.1MB

Difference: Jest has **~5.4x greater total dependencies, and ~6x greater dependency file weight**

---

## Conclusions

To accomplish the same testing objectives, do you want to:

- Install **5 - 6x / 60 MB more dependencies** (and have their related bug surface)?
- Have your tests run **7x slower, utilizing >10x more compute**?
- Be required to write **>50% more test files and code**?

And that's to say nothing of the additional bug surface that Jest brings to the table with its non-standard Node.js global mutations, which by all accounts should be a non-starter for any testing harness.

If so, please continue using Jest! Just preferably not on this project.


camjc:
What’s the speed delta?

ryanblock:
@camjc oh do not worry my human, this will all be quantified.

package.json
```diff
@@ -19,7 +19,9 @@
     "lint": "eslint .",
     "start": "node src/shared/cli/index.js",
     "qa": "pta ./src/qa/*.qa.js -r log | node src/qa/utils/qaReporter.js",
-    "test": "FILES_MODIFIED=$(git diff --name-only HEAD) jest",
+    "test": "npm run test:jest && npm run test:tape",
+    "test:tape": "tape tests/**/*-test.js | tap-spec",
+    "test:jest": "FILES_MODIFIED=$(git diff --name-only HEAD) jest",
```
qgolsteyn: Let's not have two test harnesses

src/shared/scrapers/__tests__/scrapers.all.test.js
```diff
@@ -3,14 +3,11 @@
...
 import { sync as glob } from 'fast-glob';
-import { readFileSync as readFile } from 'fs';
-import path from 'path';
 import join from '../../lib/join.js';
 import { readJSON } from '../../lib/fs.js';
-import { get } from '../../lib/fetch/get.js';
 import runScraper from './run-scraper.js';
 
-jest.mock('../../lib/fetch/get.js');
+// jest.mock('../../lib/fetch/get.js');
```
qgolsteyn: Remove comment

package.json
```diff
@@ -19,7 +19,9 @@
     "lint": "eslint .",
     "start": "node src/shared/cli/index.js",
     "qa": "pta ./src/qa/*.qa.js -r log | node src/qa/utils/qaReporter.js",
-    "test": "FILES_MODIFIED=$(git diff --name-only HEAD) jest",
+    "test": "npm run test:jest && npm run test:tape",
+    "test:tape": "tape tests/**/*-test.js | tap-spec",
```
qgolsteyn: We capture modified files for Jest in order to test only the scrapers that have changed. Is this supported by this new test harness?

qgolsteyn:
```suggestion
    "test:tape": "tape -r esm tests/**/*-test.js | tap-spec",
```

This avoids having to import `esm` in every test file

tests/integration/scrapers/scrapers-all-test.js
```diff
@@ -0,0 +1,94 @@
...
+    process.env.OVERRIDE_CACHE_PATH = testDir;
+
+    // dynamically import the scraper
+    // eslint-disable-next-line
+    const scraperObj = imports(join(testDir, '..', 'index.js'));
+
+    if (scraperObj.state === 'AL' && scraperObj.country === 'USA') {
+      // Honestly these linter rules are absurd
+      // eslint-disable-next-line
+      scraperObj.scraper = scraperObj.scraper[0];
```
qgolsteyn: This should be removed

tests/integration/scrapers/scrapers-new-test.js
```diff
@@ -0,0 +1,36 @@
...
+const exec = require('child_process').execSync;
+
+const shared = join(process.cwd(), 'src', 'shared');
+const lib = join(shared, 'lib');
+
+const fs = imports(join(lib, 'fs.js'));
+const schema = imports(join(lib, 'schema.js'));
+const runScraper = imports('./run-scraper.js').default;
+
+const command = 'git diff --name-only HEAD';
```
qgolsteyn: We use different diffing strategies depending whether we are running for a PR (diff against master) or just a regular commit (diff against previous commit). Would recommend keeping the env variable `FILE_MODIFIED` which we set in the `package.json` scripts

tests/unit/shared/lib/fs-test.js
```diff
@@ -0,0 +1,224 @@
...
+const fs = imports(join(process.cwd(), 'src', 'shared', 'lib', 'fs.js'));
+
+test('Module exists', t => {
+  t.plan(1);
+  t.ok(fs, 'fs exists');
+});
+
+test('fs.exists', async t => {
+  const tests = ['test/somefile', 'test/asubdirectory/somefile', 'somefile'];
+  t.plan(tests.length * 3);
```
qgolsteyn: I see having to plan tests as being a big source of headaches for someone not familiar with this library. Is there a way to avoid having to specify the number of tests? Or could we document this?

package.json
ryanblock: Of course. This is so folks could test things out.

src/shared/scrapers/__tests__/scrapers.all.test.js
ryanblock: I had to change how that test worked in order to port this test without Jest's mocking; that comment needs to remain in place for this test to continue working.

package.json
ryanblock: Please see the individual ported test in question; this is a responsibility of the test. Putting it into an env var is very hacky.

ryanblock: I'm not sure what side effects that will have, and I don't believe it's necessary at this time.

tests/integration/scrapers/scrapers-all-test.js
ryanblock: Agree – this test is flawed. Re. scope: I did not set out to improve these tests, I set out to port them.

tests/unit/shared/lib/fs-test.js
ryanblock: Can document. Will explain why you should plan your tests in comments above.

tests/integration/scrapers/scrapers-new-test.js
ryanblock: I fail to see how the result differs between running it via the npm script and running it within the test, can you elucidate?

qgolsteyn: See `.github/workflows/test.yml`, we run `git diff --name-only origin/master` when running the workflow in a PR as we want to compare against master. This test will need to do the same to avoid a regression.

package.json
qgolsteyn: See my comment for the test in question, we need to diff differently depending on whether we are in a PR or performing regular testing

ryanblock:
@qgolsteyn

> t.plan() is definitely a sticking point for me. Is there a way to avoid it? If not, could we document it somewhere for someone who wants to add tests.

This is legit; I used to be pretty annoyed about `t.plan()` early on in my testing career. Then I saw the light.

Good tests are about **determinism**: getting **exactly** the results you're seeking **every** time. (This is why harnesses that rewrite runtime globals are a fatal mistake.)

A test that is authored incorrectly could report back incorrect results that without a plan would not be caught. For example:
- Say you're testing a code path three times
- Now say your code returns two (or four) times
- With `t.end()`, your test would pass; with `t.plan(2)` (or `4`), your tests would fail

Another example:
- Say one is testing an asynchronous code path
- Async in js is a frequent bug vector; say an incorrectly authored calls `t.end()` before execution completed
- Your tests would pass and you may not notice that your actual code path didn't return

That is not a contrived example; I have seen (and actually made) these mistakes commonly. Any test harness that does not have a plan for a known number of results is, in my opinion, not truly deterministic.

This looks like an annoyance or a bug, but in reality is a powerful feature and should always be employed.

> We can tell tape to use esm when importing test files, avoiding the need to import esm in every test file.

I'm not convinced we should, but that's a very minor implementation detail that I don't think factors into any decisions around Jest.

> Could we get rid of Jest entirely? No need to have two test frameworks

Of course! Not proposing we keep both, kept them in the PR so if anyone one would like to replicate my results, they can do so without having multiple branches/repos running at the same time.

tests/integration/scrapers/scrapers-new-test.js
ryanblock: Got it, was not aware.

So in CI we run `git diff --name-only origin/master`, while locally we run `git diff --name-only HEAD`, yes? 

Will memorialize in the test.

lazd:
+1 for tape, -1 for jest, let's merge this and move the last set of tests to tape and remove jest completely.

PR #486, opened 2020-03-29
Fix some lint

camjc:
Found using LGTM https://lgtm.com/projects/g/lazd/coronadatascraper/alerts/?mode=list

*Please review per commit*

There are still a few in find-features
image
but that function is pretty unwieldy so feel like we might want to simplify before removing those.

### DIFF
--- a/site/crossCheckReport.js
+++ b/site/crossCheckReport.js
@@ -35,8 +35,8 @@ function crossCheckReportTemplate(report) {
             <tr>
     `;
     const sourceURLShort = source.url.match(/^(?:https?:\/\/)?(?:[^@/\n]+@)?(?:www\.)?([^:/?\n]+)/)[1];
-    const curators = getContributors(source.curators, 'Curated by');
-    const sources = getContributors(source.sources, 'Sourced from');
+    const curators = getContributors(source.curators);
+    const sources = getContributors(source.sources);
     html += `<th class="cds-SourceComparison-source">`;
     if (source.curators) {
       html += `<strong>${curators}</strong>`;
--- a/site/lib/geography.js
+++ b/site/lib/geography.js
@@ -1,27 +1,8 @@
-/*
-  Get the full name of a location
-*/
-export const getName = function(location) {
-  let name = '';
-  let sep = '';
-  if (location.city) {
-    name += location.city;
-    sep = ', ';
-  }
-  if (location.county) {
-    name += sep + location.county;
-    sep = ', ';
-  }
-  if (location.state) {
-    name += sep + location.state;
-    sep = ', ';
-  }
-  if (location.country) {
-    name += sep + location.country;
-    sep = ', ';
-  }
-  return name;
-};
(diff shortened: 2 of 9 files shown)


site/crossCheckReport.js
```diff
@@ -35,8 +35,8 @@ function crossCheckReportTemplate(report) {
             <tr>
     `;
     const sourceURLShort = source.url.match(/^(?:https?:\/\/)?(?:[^@/\n]+@)?(?:www\.)?([^:/?\n]+)/)[1];
-    const curators = getContributors(source.curators, 'Curated by');
-    const sources = getContributors(source.sources, 'Sourced from');
```
lazd: Not lint, legit error -- looks like I screwed it up when I refactored.

site/lib/geography.js
```diff
@@ -1,27 +1,8 @@
...
-    name += sep + location.country;
-    sep = ', ';
-  }
-  return name;
-};
+/** Get the full name of a location
+ * @param {{ city: string?; county: string?; state: string?; country: string?; }} location
+ */
+export const getName = location =>
+  [location.city, location.county, location.state, location.country].filter(Boolean).join(', ');
```
lazd: nice

src/events/processor/find-features/index.js
```diff
@@ -104,6 +104,16 @@ function matchFeature(props, featuresData) {
...
 
+/**
+ * @param {string?} locationItem
+ * @param {{ properties: { name: string?; name_en: string?; region: string?; }; }} feature
+ */
+const doesLocationMatchFeature = (locationItem, feature) =>
+  locationItem &&
+  [feature.properties.name, feature.properties.name_en, feature.properties.region].some(
+    toMatch => toMatch === locationItem
+  );
```
lazd: Nice, but I personally would rather see this as a proper `function` with `return` just so my brain doesn't hurt.

src/shared/scrapers/USA/NYT.js
```diff
@@ -44,10 +44,6 @@ const scraper = {
       throw new Error(`Timeseries starts later than SCRAPE_DATE ${scrapeDateString}`);
     }
 
-    if (scrapeDate < firstDateInTimeseries) {
-      throw new Error(`Timeseries starts later than SCRAPE_DATE ${scrapeDateString}`);
-    }
```
lazd: Never hurts to check twice, right? :P

src/events/processor/find-features/index.js
lazd:
```suggestion
function doesLocationMatchFeature(locationItem, feature) {
  return locationItem &&
  [feature.properties.name, feature.properties.name_en, feature.properties.region].some(
    toMatch => toMatch === locationItem
  );
}
```

camjc: Why?
Assuming a good reason, we can enforce via linting if we want to.

PR #494, opened 2020-03-29
Added scraper for Nova Scotia, Canada

chunder:


src/shared/scrapers/CAN/NS/index.js
```diff
@@ -0,0 +1,64 @@
...
+        const pos = parse.number(row.Positive);
+        const neg = parse.number(row.Negative);
+
+        return {
+          cases: pos,
+          tested: pos + neg
+        };
+      }
+    }
+  }
```
lazd: Not tested, but this should handle the case that the timeseries does not contain a sample for that date (i.e. it's missing one in the middle):
```suggestion
    throw new Error(`Timeseries does not contain a sample for SCRAPE_DATE ${scrapeDateString}`);
  }
```

Issue #490, opened 2020-03-29
Migrate away from needle


hyperknot:
Yesterday I've spent an hour trying to figure out the particular behaviour of needle, but I couldn't. I think it's just behaving in a very unreliable way.

Doing nothing but changing the URL between smaller (1 KB) and bigger (10 MB) files, it was producing the following behaviour:
- tiny files were handled always correctly
- 10 MB files were either:
   - downloaded (the least probable one)
   - simply discarded, even though the success callback was called!
   - hanging the whole process and not letting it finish at the very end, for like 30 seconds, so it couldn't terminate

At the end I tried a bunch of other libraries and all of them worked correctly 100% of the time. So I guess this needle has to be seriously unreliable.

I looked at alternatives from this issue:
https://github.com/request/request/issues/3143

So far I really like [got](https://github.com/sindresorhus/got), modern library, written with native async/await style in mind. 

Also, saving a file to disk with this is literally 2 lines, and is one of the first examples in the docs. Many other libraries don't even mention the streaming file saving in their docs at all.

Here is a comparison on their page:
https://github.com/sindresorhus/got#comparison

I recommend migrating from needle to got.

lazd:
I'd like to try it -- sometimes needle fails to download Utah for me, stopping half way through...

chunder:
I also just spent time debugging the fact that needle doesn't download UT's 11MB payload reliably. I also made a PR to have it retry.

Using got sound better to me.

ryanblock:
Example source(s) that have these large payloads so that I can test out some alternatives?

chunder:
UT, USA.

jzohrab:
Hey @ryanblock, @hyperknot , @lazd , needle does let you stream large docs, ref https://github.com/tomas/needle#get-a-very-large-document-in-a-stream-from-07

I just downloaded a 20 MB PDF from the interwebs:
```
const imports = require('esm')(module);
const needle = imports('needle');
const fs = imports('fs');

// This file is 20 MB
const url = "https://www.hq.nasa.gov/alsj/a17/A17_FlightPlan.pdf"

let opts = { content_type: 'application/pdf' };
var stream = needle.get(url, opts);
let out = fs.createWriteStream('got.pdf');

stream
  .pipe(out)
  .on('readable', function() {
    var chunk;
    while (chunk = this.read()) {
      console.log('got data: ', chunk);
    }
  });
```

Yes, it took a while (my bandwidth here is bad), but no issues.

[...]

PR #496, opened 2020-03-29
Check that bytes received equals the Content-Length header; retry if not

chunder:


lazd:
Can we explore using `got` instead? It seems more reliable, see @hyperknot's comments on #490

lazd:
I'm gonna merge this, but we should look into `got` ASAP!

PR #501, opened 2020-03-29
fix: DE stale data

cmcjacob:


lazd:
Thanks!

lazd:
Was missing the rollup, no biggie. Thanks!

PR #503, opened 2020-03-29
Added Puerto Rico.

jacobmcgowan:
I initially tried to add PR under USA like GU was; however, the feature could not be found for PR, USA. I added it as a country instead.

### DIFF
--- a/src/shared/scrapers/PRI/index.js
+++ b/src/shared/scrapers/PRI/index.js
@@ -19,7 +19,6 @@ const scraper = {
     }
   ],
   async scraper() {
-    console.log('test');
     const $ = await fetch.page(this.url);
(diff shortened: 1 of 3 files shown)


lazd:
Weird that it can't be found under USA...

jacobmcgowan:
I believe it has something to due with the `woe_label`, `woe_name`, `admin`, `geonunit`, and similar fields in world-states-provinces.json file. For states, `admin` is "United States of America" while for PR it is "Puerto Rico" for example.

lazd:
Hmmm, I think PRI makes more sense as its own country, given the ISO 3166 alpha-3 code, despite it being a US territory. I'll merge as-is!

Issue #49, opened 2020-03-15
Document output data format


lazd:
We need a nice rundown of each field, what it is, what format its in, etc.

DavidGeeraerts:
Yes that's me [jumping up and down]!

DavidGeeraerts:
I'm looking into whether to make this an [ISO 19115:2013](https://en.wikipedia.org/wiki/Geospatial_metadatal) metadata record or use [metadata-json](https://project-open-data.cio.gov/v1.1/schema/). 
This is after all, Vector Spatial Data --> ISO 19115:2013

lazd:
@hyperknot, thoughts on this?

DavidGeeraerts:
I'm heavily swayed by data.gov Open Data Standards:
[Project Open Data Metadata Schema ](v1.1https://project-open-data.cio.gov/v1.1/schema/)

I'm somewhat spinning my wheels trying to find a good metadata-json editor. I'd rather not have to edit in raw. I'm looking...

DavidGeeraerts:
This is exactly what I was looking for and need!
[mdEditor](https://go.mdeditor.org/#/dashboard)
Let's just say it's in the works after being stalled for the day. :-(

DavidGeeraerts:
@lazd I've validated the workflow with [mdEditor](https://www.mdeditor.org/l):

1. Import the mdeditor metadata file
2. merge
3. edit
4. export
Think this workflow will work as a collaborative metadata project.

My question: should I upload the metadata.json file to the coronadatascraper repository?
I would think so, just want to check with you, since that's not exactly where the data is.
Should another repository be created just for the data?
(organization: CoronaVirusDataTeam)

lazd:
@DavidGeeraerts how often will this metadata file change, what does it look like, and how should we provide it?

DavidGeeraerts:
Getting an official metadata record for data.gov is going to take (me) a while. Went back to creating the metadata.md file so you can post on [your] website. I've started a Github organization called ["CoronaVirusDataTeam"](https://github.com/CoronaVirusDataTeam) --that was before the whole Coronavirus Response Data Team)-- and have a repo for docs. In there is the draft for [metadata.md](https://github.com/CoronaVirusDataTeam/docs/blob/master/metadata.md). This is a lot of work and I can only do it in chunks as time permits.

Can you provide me with (or update the [document](https://github.com/CoronaVirusDataTeam/docs)):

featureId:
   description: <needed>

[...]

PR #505, opened 2020-03-29
separate documentation for data fields

piccolbo:
This is intended as a user-facing document, not a developer one.

### DIFF
--- a/docs/data_fields.md
+++ b/docs/data_fields.md
@@ -47,11 +47,10 @@ The following fields detail the data sources:
 * `maintainers` - An array of maintainers responsible for writing the scraper code that obtains the data
 
 
-### `timeseries.csv`
+### `timeseries-tidy.csv`
 
 For each entry, the following data is provided:
 
 * `date` - the date the data point refers to
 * `type` - the type of data point: cases, tested, deaths, hospitalized, discharged, or recovered
 * `value` - the value of the data point (a cumulative count of events of a certain type)
-
--- a/docs/data_fields.md
+++ b/docs/data_fields.md
@@ -1,28 +1,57 @@
+# Data fields
 
-The main fields are the following, plus the geographical entity fields explained below:
+## All files
 
-* date: the date the data point refers to
-* type: the type of data point: cases, tested, deaths, hospitalized, discharged, or recovered
-* value: the value of the data point (a cumulative count of events of a certain type)
+The following fields are available in all files for each geographical entity:
 
-Each observation is associated with a geographical entity, identified in the following four fields.
+* `name` - the full name of the geographical entity being represented
+* `city` - the name of the city
+* `county` - county or parish or the appropriate name of the administrative subdivision below the level of state or equivalent 
+* `state` - state, province or region depending on jurisdiction. In general, the first administrative subdivision below the level of country.
+* `country` - ISO 3166-1 alpha-3 (three letter) country code 
+* `level` - one of `city`, `county`, `state`, `country`. Provided in order to facilitate filtering
 
+The `country` field is mandatory. In general, whenever the record is about administrative subdivisions of a level, you'll find all larger levels are non-empty.
 
-* city: the name of the city
-* county: county or parish or the appropriate name of the administrative subdivision below the level of state or equivalent 
-* state: state, province or region depending on jurisdiction. In general, the first administrative subdivision below the level of country.
-* country:  ISO 3166-1 alpha-3 (three letter) country code 
-
-The country field is mandatory. In general, whenever the record is about administrative subdivisions of a level, al including levels are non empty. One exception is New York City because it is subdivided into five counties.
+However, there are exceptions. New York City will not have a `county` field because it is subdivided into five counties.
 
 The following fields are uniquely determined by the geographical entity and are provided as a convenience.
 
-* population: a recent estimate of the population in the geographical entity
-* lat: latitude of the geographical entity
-* long: longitude of the geographical entity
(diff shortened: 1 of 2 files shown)


docs/data_fields.md
```diff
@@ -0,0 +1,28 @@
+
+The main fields are the following, plus the geographical entity fields explained below:
+
+* date: the date the data point refers to
```
piccolbo: without a time offset, the date is ambiguous. Unfortunately , a time offset is not unambiguously computable from a time zone.

docs/data_fields.md
```diff
@@ -0,0 +1,28 @@
...
+* type: the type of data point: cases, tested, deaths, hospitalized, discharged, or recovered
```
piccolbo: maybe more detail on what these words mean exactly

docs/data_fields.md
```diff
@@ -0,0 +1,28 @@
...
+* value: the value of the data point (a count of events of a certain type)
+
+Each observation is associated with a geographical entity, identified in the following four fields.
+
+
+* city: the name of the city
+* county: county or parish or the appropriate name of the administrative subdivision below the level of state or equivalent 
+* state: state, province or region depending on jurisdiction. In general, the first administrative subdivision below the level of country.
```
piccolbo: Unlike the country, we don't refer to a standard for these names. This can make joining other datasets difficult, particularly for counties in US and maybe elsewhere.

docs/data_fields.md
```diff
@@ -0,0 +1,28 @@
...
+* country:  ISO 3166-1 alpha-3 (three letter) country code 
+
+The country field is mandatory. In general, whenever the record is about administrative subdivisions of a level, al including levels are non empty. One exception is New York City because it is subdivided into five counties.
```
piccolbo: This sounds open to ambiguity.

docs/data_fields.md
```diff
@@ -0,0 +1,28 @@
...
+
+The following fields are uniquely determined by the geographical entity and are provided as a convenience.
+
+* population: a recent estimate of the population in the geographical entity
```
piccolbo: Do we know more about the source of the pop estimate

docs/data_fields.md
```diff
@@ -0,0 +1,28 @@
...
+* lat: latitude of the geographical entity
+* long: longitude of the geographical entity
```
piccolbo: Which specific point are we producing?

docs/data_fields.md
```diff
@@ -0,0 +1,28 @@
...
+* tz: time zone of the geographical entity
```
piccolbo: some geographical entries include multiple time zones (US, Russia, China and more)

docs/data_fields.md
camjc: especially that they're all cumulative, and that tested means regardless of positive/negative result.

piccolbo: added the word "cumulative" (good point). I would like to know a bit more for each type though, but let's not the perfect get in the way of the good.

camjc:
Do we want to link to this from somewhere? Is there existing documentation that should be cleaned up?

piccolbo:
Yes I think the idea was to link from the README. Do you think some of my questions can be addressed by someone or is this a case of the perfect getting in the way of the good?

docs/data_fields.md
lazd: Yes, it will come from https://github.com/hyperknot/country-levels and other census sources we have in the repo.

docs/data_fields.md
lazd: That's true. We can align around something @hyperknot

docs/data_fields.md
piccolbo: I was thinking more about original source, like is it world bank, UN, US census etc, what year and so on. The exact file we keep it in is of little consequence for a data consumer.

docs/data_fields.md
piccolbo: I see there is discussion going on as to the correct approach. In the interest of delivering this document, I will make an issue to document this after the new standards are in place

piccolbo: I think this is premature while discussion is ongoing and out of scope for this PR (We can't document what is not done yet, and it's my fault for mentioning it here). I entered #601 to remind ourselves of this task. Please mention it when work on this starts.

lazd:
@camjc README link added

docs/data_fields.md
```diff
@@ -0,0 +1,57 @@
...
+
+The following fields detail the data sources:
+
+* `url` - The exact URL from which the data was obtained
+* `sources` - An array of sources that published the data
+* `curators` - An array of curators responsible for manually curating the data
+* `maintainers` - An array of maintainers responsible for writing the scraper code that obtains the data
+
+
+### `timeseries.csv`
```
lazd, line 50:
```suggestion
### `timeseries-tidy.csv`
```

lazd:
Still working on this one, I hope to have it done today.

DavidGeeraerts:
@piccolbo This has been in the works, but I stopped because the data schema was WAY TOO VOLATILE! Once the data schema solidifies, I'm hoping to get back to it:
[metadata.md](https://github.com/CoronaVirusDataTeam/docs/blob/master/metadata.md)

piccolbo:
I noticed #49. I asked Larry to do something quick and in English. He gave
me the ok. I don't think schema evolution is a good reason to go without
documentation. I don't understand much of #49 but it seems the output would
be metadata not English. So I don't see overlap.

PR #507, opened 2020-03-30
Add html table validator class, add to PA scraper

jzohrab:
A simple HTML table "schema validator" with tests and docs.

In summary, the user creates a hash of rules the table should follow, and validates the table against that:

```
const rules = {
  headings: {
    0: /country/i,
    1: /number of cases/i,
  },
  data: [
    { column: 0, row: 'ANY', rule: /Adams/ },
    { column: 1, row: 'ALL', rule: /^[0-9]+$/ },
  ]
};
HtmlTableValidor.throwIfErrors(rules, $table);
```

Data rules can match 'ANY', 'ALL', or a single row.

If the `throwsIfErrors` call fails, it (optionally) logs to console, and throws an exception with the validation failures.  Sample failure strings (pulled from the unit tests):

* for a heading rule: `'heading 0 "location" does not match /something/'`
* for a data ANY rule: `'no row in column 17 matches /^[0-9]+$/'`
* for a data ALL rule: `"deer county" in column 0 does not match /apple/`
* for a data row = number rule: `cell[2, 1] value "66" does not match /cat/`

An example is in the docstring and in docs/sources.md.

The validator really only works for simple tables, i.e. TRs with lots of TDs, and not nested tables, but it would work at least for the tables I've seen!  I believe it could handle other more complex tables, such as tables with section/subtotal rows, with some changes.  Haven't really thought about it, and I didn't try to overgeneralize this first iteration.

I added this validator to the PA scraper as a live example, but can revert that commit and repush if you want to keep it out.  Performance should be reasonable, as we're just doing indexing lookups and regexing, but I don't know how to load test this just yet.  At any rate, this check shouldn't be less performant than many of the HTML scrapers.

I'm happy to take suggestions for code style, docs, etc, but hope those can be added as subsequent PRs.  It would be nice to get this work into the codebase, even as a dark release (with the PA usage commented out).  Cheers!


lazd:
@ryanblock @qgolsteyn @chunder @hyperknot let's get a CR going on this PR, it could be good stuff.

src/shared/lib/html-table-validator.js
```diff
@@ -0,0 +1,289 @@
...
+    }
+    throw new Error(`${msg}.  Sample: ${firstN.join(';')}.`);
+  }
+
+  // PRIVATE (ish) //////////////////////////////////////////////
+  //
+  // Javascript doesn't really have private, prepending underscore
+  // to discourage direct calls.
+  //
+  // Most of these methods should be static b/c they don't use "this",
```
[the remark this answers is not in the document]
jzohrab: It's not really "broken" :-) , but I wasn't sure how to handle this.  These are helper functions only.  I found it code clutter to say `HtmlTableValidator.my_func()` instead of `my_func()`, but the linter insisted on the longer name.  Since this was lint only, I figured hiding the warning was ok.

ps - This is me not knowing javascript, if there's a better way to do it, LMK!

src/shared/scrapers/USA/PA/index.js
```diff
@@ -179,6 +180,22 @@ const scraper = {
...
+          2: /deaths/i
+        },
+        data: [
+          { column: 0, row: 'ANY', rule: /Adams/ },
+          { column: 1, row: 'ALL', rule: /^[0-9]+$/ },
+          { column: 2, row: 'ALL', rule: /(^[0-9]+|)$/ }
+        ]
+      };
+      const opts = { includeErrCount: 5, logToConsole: true };
+      HtmlTableValidator.throwIfErrors(rules, $countyTable, opts);
```
qgolsteyn, line 197: Based on the usage I'm seeing, we should rewrite the class as a function. This will simplify the code, and allow you to specify which functions can be imported in other parts of the codebase

jzohrab: That change would be quick.  If that’s the team preference can do.  That public static was just a helper func to simplify the API and usage, but the class itself has hardly any state IIRC.

jzohrab: I'm going to assume that this is what the team wants and will make this change, thanks @qgolsteyn.

qgolsteyn: Let me know once you are done this change and I'm happy to review again

chunder:
Related idea: It would be cool to give the `rules` structure and the root cheerio object `$` to a function... and have it *find* the table that obeys the rules. Maybe even feed it through `tableparser` too. If not table matches the rules, throw or return null or something.

chunder:
You could possibly give it a subtree of the DOM instead of the root `$` if you need to scope it down first. But if you flesh out the rules, only 1 table should match them.

jzohrab:
@chunder - I had the same thought after opening this PR ... this PR is a good step in that direction.  I think it could easily find a table if the rules were explicit, and then go from there.  It could even return a fully populated dataset.  So let's get this in as a first iteration, and then if it's useful and a net gain we can expand on the idea.

src/shared/scrapers/USA/PA/index.js
jzohrab: Pushed @qgolsteyn, thank you!  I only exported a few methods to keep the module API as clean as possible, others can export things if they need.

camjc: @jzohrab are we still migrating from a class to a function?

jzohrab: @camjc that snippet is outdated, everything is a function now.  The code now reads

```
import * as htmlTableValidation from '../../../lib/html/table-validation.js';
...
htmlTableValidator.throwIfErrors(rules, $countyTable, opts);
```

PR #509, opened 2020-03-30
Folder from zip url, downloadFile functions

hyperknot:
Added a [got](https://github.com/sindresorhus/got) based downloadFile function, that's the easy bit.

The complex one is a caching download zip + extract into folder function. This is needed because we won't be able to add bigger dependencies because of lambda. Also I don't think it's a good idea to put big GitHub repositories containing thousands of GeoJSON files into a package.json. I tried, it worked but it really made yarn slow.

Downloading and unzipping small zips from GitHub is a much better solution I believe. The downloads are cached by using the __cache.__json to store the original URL.


package.json
```diff
@@ -37,15 +37,16 @@
     "@turf/union": "^6.2.0-alpha.1",
     "ajv": "^6.12.0",
     "cheerio": "^1.0.0-rc.3",
-    "cheerio-tableparser": "^1.0.1",
```
ryanblock: This is probably your cheerio-tableparser issue

hyperknot:
I have no idea why is this breaking the `cheerio-tableparser` imports. I have only included 2 new packages, haven't touched anything else

Also I tried rm yarn.lock and recreating it, but the cheerio stayed the same. If someone could figure out what is happening with cheerio-tableparser it'd be very helpful.

ryanblock:
Can we zoom further out here: what are we trying to accomplish for what use case, and what are the requirements?

hyperknot:
> Can we zoom further out here: what are we trying to accomplish for what use case, and what are the requirements?

I want to be able to include some bigger assets in vendor. For example one is 70 MB. Also there is a Brazil county PR which I think will also be quite big, it looks very detailed.

So I'm trying to find a way to include those big assets without having the package explode in size.

hyperknot:
Basically this is my best solution for https://github.com/lazd/coronadatascraper/issues/489

ryanblock:
Ah, ok! So you want to dynamically operate on a very large static dataset.

My suggestion will be to do this via S3 – having it in-network is going to be much faster than going over the internet, and we'll be able to ensure its availability.

Is this a thing you need right now in order to proceed on a certain dataset?

hyperknot:
I mean this would work right now as it is. Later on we can just change the target folder to S3, if we mount it via some kind of file system mount. How does lambda mount S3?

70 MB is not very large, I believe by the time we have counties from all countries we'll be in the hundreds of MBs range.

ryanblock:
We'll load out of S3 via the AWS SDK, which sustains transfers at many GBps.

hyperknot:
I guess that will come as a higher layer, saving and restoring a whole folder for example, invisible for the lib functions, wouldn't it? So we can merge this one now and just change the target folder once the S3 integration is done, can't we?

hyperknot:
This is now not used, turns out yarn has a reliable way of using remote .tgz files.

PR #510, opened 2020-03-30
RFC: Add tesseract and kern county scraper

livejake:
Need to suppress the error messages and know that eng.traineddata isn't in an ideal place, but just pushing it up for comments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "needle": "^2.3.3",
     "pdfreader": "^1.0.6",
     "puppeteer": "^2.1.1",
+    "tesseract.js": "^2.1.1",
     "yargs": "^15.3.1"
   },
   "devDependencies": {
--- a/src/shared/lib/fetch/index.js
+++ b/src/shared/lib/fetch/index.js
@@ -2,6 +2,7 @@ import cheerio from 'cheerio';
 import csvParse from 'csv-parse';
 import { PdfReader } from 'pdfreader';
 import puppeteer from 'puppeteer';
+import Tesseract from 'tesseract.js';
 import * as caching from './caching.js';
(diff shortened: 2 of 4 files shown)


hyperknot:
How big is the whole tesseract package + the OCR model for it?

livejake:
TesseractJS is 1.1 MB, the model eng.traineddata is 23.1 mb

hyperknot:
Also affected here: https://github.com/lazd/coronadatascraper/issues/489

livejake:
Could maybe use something like https://github.com/shelfio/aws-lambda-tesseract to get size down to 6 mb

camjc:
Do we really want to head down the path of OCR?
Seems like data reliability would massively decrease if we use OCR? CC @lazd @ryanblock

livejake:
If OCR is the only option, then I think it can be a valuable resource. However, given its reliability issues, we should require additional testing. For example, in this PR, I test for internal consistency. I could also look at if text strings changed or some other attributes.

jzohrab:
Hi @livejake , are you continuing work on this PR's branch?  Cheers! jz

lazd:
TBH, I don't think scraping images is really within scope. I'd rather find a person in a given state to curate a data source for us until the state/county figures out how to internet. In this case, Harriet Rowan curates CA for us daily and does a fantastic job. Though this is definitely a magnificent application of the technology and it's impressive that it's been made to work, it's not sustainable and would be potentially quite brittle (like our scrapers aren't already, heh).

So I say we close this, but tip our hats to @livejake for attempting to work with the absolute boneheads at the Kern County Public Health Services Department ❤️

livejake:
Ok. I understand your perspective. If you ever change your mind, happy to help finalize.

PR #511, opened 2020-03-30
Add US Virgin Islands

ParkerM:
This PR adds a scraper for the VI department of health page in the google sheet. The location is funky because the alpha-3 code is VIR but it's also listed in `usa-states.json` as VI. I left it as VI USA for now since I wasn't sure how it should be classified.


src/shared/scrapers/USA/VI/index.js
```diff
@@ -0,0 +1,51 @@
+import * as fetch from '../../../lib/fetch/index.js';
+import * as parse from '../../../lib/parse.js';
+
+const scraper = {
+  state: 'VI',
+  country: 'USA', // The alpha-3 country code is VIR
```
lazd: Hmm, then do we need to have `VIR` here instead?

src/shared/scrapers/USA/VI/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+    // Date format: "Sunday, March 29, 2020 7:00 pm"
+    // const lastUpdated = data.get('last updated');
+
+    const stats = {
+      cases: positive,
+      tested: totalTested
+    };
+
+    // Need to wrap this in an array or else the test fails
```
lazd: @qgolsteyn test's shouldn't fail if an array isn't returned, what's the deal there?

src/shared/scrapers/USA/VI/index.js
ParkerM: Your call - I tried a few different combinations of country/state/featureId but couldn't find one that was matched during the update task.

src/shared/scrapers/USA/VI/index.js
qgolsteyn: Interesting, let me look into it

ParkerM: For reference, the issue was an unhandled promise rejection error being thrown [here](https://github.com/covidatlas/coronadatascraper/blob/f4b6d4c68bc5b43d05fe39148d2b573c97b64ecb/tests/integration/scrapers/scrapers-all-test.js#L86) when I returned just the object instead of the array wrapper.

qgolsteyn: I could not replicate, if I give the updated scraper an object instead of an array, it works fine

lazd:
Thanks, @ParkerM !

PR #512, opened 2020-03-30
fix AUS ACT (deaths) and consolidate key lookup

camjc:
Trying out some patterns for how best to lookup. Wondering if there can be a shared lookup across scrapers (like if the word deaths is included, then its very probable to be the deaths item).

AUS ACT has a death now, so they've added 'lives lost'.


lazd:
This looks fine to me, the shared code is fine too. Is this ready to go in?

camjc:
Yes please :)

PR #517, opened 2020-03-30
ISO dates, round 2

HerbCaudill:
This PR is another shot at solving the problems addressed in #393. 

Rather than introduce a separate module of date functions, this adds the new ones to the existing `datetime` library and changes out the internals. 

Unit tests are added covering all of the `datetime` library's functionality. All tests have been ported to tape.

## Work in progress

I've opened this as a draft for review. There is some additional work to be done to non-ISO dates from the system. 

Also, at this point I'm not confident that this won't break existing functionality. Specifically: All of the old date functions ultimately depended on `new Date(...)` to parse dates in a much more permissive (and timezone-unsafe) way. The new functions use `datetime.parse(...)` to parse dates, which accepts a limited set of formats. (Alternative delimiters such as slashes and dots are accepted as well.)
```
YYYY-MM-DD
YYYY-M-D
M-D-YYYY
M-D-YY 
```

### TODO

- [ ] Replace all direct references to `process.env.SCRAPE_DATE`
- [ ] Replace all direct usage of the JS `Date` object
- [ ] Update cache mechanism to use ISO dates in paths
- [ ] Update existing cache folder names


src/shared/lib/datetime.js
```diff
@@ -1,106 +1,146 @@
...
-  let localDate = date;
-  if (typeof localDate === 'string') {
-    localDate = new Date(localDate);
-  }
-  const month = localDate.getUTCMonth() + 1;
-  const day = localDate.getUTCDate();
-  const year = localDate.getUTCFullYear();
+// util functions
+
+const currentJsDate = () => new Date(Date.now()); // allows us to mock current date
```
hyperknot: I recommend getting the `.now()` at UTC timezone.

hyperknot: > allows us to mock current date

Or is this only used for testing? Can we put testing related functions to a test prefixed file?

src/shared/lib/datetime.js
```diff
@@ -1,106 +1,146 @@
...
+
+const currentZdt = () => ZonedDateTime.parse(currentJsDate().toISOString());
 
-  return `${year}${sep}${month}${sep}${day}`;
+const normalize = d => d.replace(/[\\/.]/g, '-'); // replaces slashes & dots with dashes
+
+// truncate ISO datetime to ISO date
+const truncate = datetime =>
+  datetime
```
hyperknot: I recommend taking the first 10 characters. It's faster and more reliable.

Also, where is space in `2020-03-30T18:26:20.708516+02:00`?

src/shared/lib/datetime.js
```diff
@@ -1,106 +1,146 @@
...
-  const day = localDate.getUTCDate();
-  const year = localDate.getUTCFullYear();
+// util functions
+
+const currentJsDate = () => new Date(Date.now()); // allows us to mock current date
+
+const currentZdt = () => ZonedDateTime.parse(currentJsDate().toISOString());
 
-  return `${year}${sep}${month}${sep}${day}`;
+const normalize = d => d.replace(/[\\/.]/g, '-'); // replaces slashes & dots with dashes
```
hyperknot: What does this do? What kind of object is d here? Can you add a comment?

hyperknot:
Thanks a lot for doing the ISO date conversion of the project, it will be a very important improvement! I added a few small notes, but otherwise looks good to me.

hyperknot:
@HerbCaudill the new JHU has the following weird formats, both of them in a single file:
US: '1/22/2020'
Global: '1/22/20'

Can you parse both?

HerbCaudill:
Yes

HerbCaudill:
Closed in favor of #541

PR #518, opened 2020-03-30
Add county level FIPS code to data output

slezakbs:
The Federal Information Processing Standard (FIPS) makes it easier to match USA counties with other datasets.


lazd:
@hyperknot thoughts?

hyperknot:
Of course we should be migrating to this as soon as my PR is merged, which is hopefully in a few hours. Basically we should use ids like `fips:06085` to refer to counties.

I've parsed the up-to-date XLSX and CSV datasets from Census, as well as the shape files from Census.
https://github.com/hyperknot/country-levels/blob/master/docs/fips_list.md

All county GeoJSON here:
https://github.com/hyperknot/country-levels/blob/master/export/geojson/q5/fips_all.geojson

slezakbs:
I had already forked and made some quick changes for me to use.  Figured I'd share in case it was useful.  @hyperknot your dataset looks more complete than mine.  Looking forward to your merge!

hyperknot:
@lazd is working on this right now. Hopefully today we'll have most of fips conversion done.

PR #520, opened 2020-03-30
fix: OK (added deaths to timeseries)

cmcjacob:


chunder:
Does this work on old dates? Has the table included a death column since we began scraping it?

PR #523, opened 2020-03-30
fix: TX deaths added to timeseries

cmcjacob:


chunder:
Does this work on old dates? Has the deaths field been there the whole time?

lazd:
Just checked old dates back to `2020-3-13` where we have cache, this one doesn't break anything and deaths start appearing `2020-3-24`

PR #527, opened 2020-03-30
Country level ID

hyperknot:
Rewrote to be as simple as possible. Now only using ISO codes internationally and FIPS codes for US counties.

Dataset repo: https://github.com/hyperknot/country-levels
ISO code list: https://github.com/hyperknot/country-levels/blob/master/docs/iso1_list.md
FIPS code list: https://github.com/hyperknot/country-levels/blob/master/docs/fips_list.md

3 possible codes:
iso1, like `iso1:AU` for Australia
iso2, like `iso2:AU-NSW` for `New South Wales`
fips, for all US counties, like `fips:06085` for Santa Clara County, CA.

Population data is provided for almost all codes. GeoJSON data is provided for everything.

I don't know how to remove:
⏳ Calculating ratings...
  ❌ "undefined" .country should NOT be longer than 3 characters

I couldn't find this. If you can help me with that I think it should be good.

How to test?
I've converted 3 locations, you can try them with:

```
yarn start -i fips:06089
yarn start -i iso2:AU-NSW
yarn start -i iso1:AU
```
I've also added the -i/--id switch to specify a location in this simple form.

TODO after this: convert every scraper to use iso1/iso2/fips codes. Then we can remove all population and feature related functions. (OK, except for US city level ones).


hyperknot:
Updated first comment

lazd:
This looks super promising. How far are we from being ready here? Is it just the `undefined` thing or do we need a strategy for converting scrapers to FIPS? Can we merge without the conversion?

Has anyone taken a look at this @qgolsteyn @camjc @chunder @samirrayani?

src/shared/lib/country-levels.js
```diff
@@ -0,0 +1,89 @@
...
+const LEVELS = ['iso1', 'iso2', 'fips'];
+const TARGET_VERSION = 'v1.0.1';
+
+const levelCache = {};
+const countryLevelsDir = path.join('src', 'shared', 'vendor', 'country-levels');
+
+export function isId(str) {
+  if (!str) return false;
+  const [level, code] = str.split(':');
+  return LEVELS.includes(level) && !!code;
```
camjc: NaBD but Boolean(code) is perhaps clearer

src/shared/scrapers/AUS/index.js
```diff
@@ -28,8 +39,14 @@ const scraper = {
       const $tr = $(tr);
       const state = parse.string($tr.find('td:first-child').text());
       const cases = parse.number($tr.find('td:last-child').text());
+
+      const stateMapped = countryLevelMap[state];
+      if (!stateMapped) {
```
camjc: NaBD but can do assert(stateMapped, message here)

camjc:
Needs some merge conflict resolution and a bit of a migration plan. Happy to chat about this if useful (Do we have to big-bang or is it easy enough to support both for a day or two while we migrate?).

src/shared/lib/country-levels.js
hyperknot: Done

src/shared/scrapers/AUS/index.js
hyperknot: Thanks, fixed it.

hyperknot:
So for migration the most important thing is that we can merge this and nothing would happen. We can gradually change the country/state/county names to the ones starting with `iso1`, `iso2`, `fips` and it wouldn't effect the other regions.

One thing we can do is to come up with a human readable names, to also include in CSV for example. Something where we concatenate `county_name, state_name` for example. I think that's the only thing which might need to be implemented, otherwise it won't "hurt" any existing location.

hyperknot:
note: the original PR I made a week ago was using a .id method and stuff like that, but I totally reverted that idea. This one is just using strings and is detecting based on starting with ['iso1:', 'iso2:', 'fips:']

hyperknot:
Updated the dataset with added name_long properties for fips regions. So this way we have 

```
    "name": "Santa Clara County",
    "name_long": "Santa Clara County, CA",
```

So we can use both for debugging. I've also added a getName function which returns name_long if present, otherwise name.

hyperknot:
Now only tests are missing + the little help I need for solving the:
"undefined" .country should NOT be longer than 3 characters line

src/shared/lib/geography/index.js
```diff
@@ -103,6 +103,10 @@ export const isCity = function(location) {
   return location && location.city;
 };
 
+export const getSmallestLocationStr = function(location) {
+  return location.city || location.county || location.state || location.country || null;
```
jzohrab, line 107: Shouldn't this throw, instead of going to null?  Optionally, if we don't even know the country, perhaps it should return "Earth".

hyperknot: Good point. I believe no magic things, so definitely not "Earth", that'd wreak havoc on the datastructure at one point, but throwing is an interesting idea.

@lazd @camjc @qgolsteyn @ryanblock can you imagine a situation where all of location.city || location.county || location.state || location.country are not defined?

camjc: Throwing is safest. Use an “assert” perhaps.
God I hope we don’t hit that case though, that’d be pretty odd.

hyperknot: I've just implemented it with assert, guess what, JHU is an illegal location.

hyperknot: The best way to keep JHU a valid "location" might be to add country = "JHU" on that one.

hyperknot: The other 2 are:
- /AUS/_shared/get-data-with-tested-negative-applied.js
- /AUS/_shared/get-key.js

This is so strange, why are these picked up as locations? For the _shared ones, we should add "_shared" them to the glob filter, shouldn't we? 

For JHU, we can solve it by adding country = '-' or country = 'JHU'. Or just using null as in the original version.

ryanblock: > can you imagine a situation where all of location.city || location.county || location.state || location.country are not defined? That would be a pretty bad scenario, so I think I'm into throwing in that case!

ryanblock:
@hyperknot I'm on this now, thanks for your patience! This is potentially super because we need to get our unique keys shored up before starting to put data into Dynamo this week.

I'm trying to run this branch but I'm seeing issues.

1. callback error:
```
yarn start -i fips:06089
npx: installed 1 in 0.823s
yarn run v1.22.4
$ node src/shared/cli/index.js -i fips:06089
⏳ Fetching scrapers...
✅ Fetched 135 scrapers!
✅ All scrapers are valid!
  🐢  Cache miss for https://www.co.shasta.ca.us/covid-19/overview at coronadatascraper-cache/2020-4-1/e7e7672cc7230ad6263bb7e31253b23e.html
  🚦  Loading data for https://www.co.shasta.ca.us/covid-19/overview from server
✏️  coronadatascraper-cache/2020-4-1/e7e7672cc7230ad6263bb7e31253b23e.html written
⏳ Calculating ratings...
  ❌ "undefined" .country should NOT be longer than 3 characters
✅ Assigned ratings for 1 sources
⏳ De-duping locations...
✅ De-duped 0 locations and found 0 crosscheck failures!
✅ Data scraped!
   - 0 cities
   - 0 states
   - 1 counties
   - 0 countries
   - 6 cases
   - 0 tested
   - 0 recovered
   - 0 deaths
   - 0 active
⏳ Generating features...
Loading Country Levels data for fips
  Downloading ZIP: https://github.com/hyperknot/country-levels/releases/download/v1.1.0/export_q7.zip
  Extraction complete
✏️  /var/folders/vm/3w3c05k91p57ns9q6y320t4h0000gn/T/yINHIh/__cache__.json written
TypeError: cb is not a function
    at /Users/ryan/git/covidatlas/coronadatascraper/node_modules/extract-zip/index.js:31:23
    at /Users/ryan/git/covidatlas/coronadatascraper/node_modules/yauzl/index.js:34:21
    at FSReqCallback.oncomplete (fs.js:146:23)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
2. file not found:
```
⮑  yarn start -i iso2:AU-NSW
npx: installed 1 in 0.814s
yarn run v1.22.4
$ node src/shared/cli/index.js -i iso2:AU-NSW
⏳ Fetching scrapers...
✅ Fetched 135 scrapers!
✅ All scrapers are valid!
  🐢  Cache miss for https://www.health.nsw.gov.au/_layouts/feed.aspx?xsl=1&web=/news&page=4ac47e14-04a9-4016-b501-65a23280e841&wp=baabf81e-a904-44f1-8d59-5f6d56519965&pageurl=/news/Pages/rss-nsw-health.aspx at coronadatascraper-cache/2020-4-1/60c86b39c8060555e08387e556d875aa.html
  🚦  Loading data for https://www.health.nsw.gov.au/_layouts/feed.aspx?xsl=1&web=/news&page=4ac47e14-04a9-4016-b501-65a23280e841&wp=baabf81e-a904-44f1-8d59-5f6d56519965&pageurl=/news/Pages/rss-nsw-health.aspx from server
✏️  coronadatascraper-cache/2020-4-1/60c86b39c8060555e08387e556d875aa.html written
  🐢  Cache miss for https://www.health.nsw.gov.au/news/Pages/20200401_00.aspx at coronadatascraper-cache/2020-4-1/30e687956bcc5a8b5d9a0311e2876434.html
  🚦  Loading data for https://www.health.nsw.gov.au/news/Pages/20200401_00.aspx from server
✏️  coronadatascraper-cache/2020-4-1/30e687956bcc5a8b5d9a0311e2876434.html written
⏳ Calculating ratings...
  ❌ "undefined" .country should NOT be longer than 3 characters
✅ Assigned ratings for 1 sources
⏳ De-duping locations...
✅ De-duped 0 locations and found 0 crosscheck failures!
✅ Data scraped!
   - 0 cities
   - 1 states
   - 0 counties
   - 0 countries
   - 2182 cases
   - 105543 tested
   - 0 recovered
   - 0 deaths
   - 0 active
⏳ Generating features...
Loading Country Levels data for iso2
(node:65239) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open 'src/shared/vendor/country-levels/iso2.json'
(node:65239) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:65239) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
✨  Done in 33.36s.
```

Here's what I'd like to see / know:
- How big are the files we need to vendor? If reasonable, we should vendor them.
- We should not do dynamic downloads right now, let's keep the change delta as small as possible here
  - We'll definitely replace needle in another PR, 100%
  - Also the large file concern should go away in the next week on its own w S3 sideloading
- Please present a clear before/after picture of scraper changes you're proposing so we can clearly understand what needs to change?

Thank you!

hyperknot:
It's not possible to vendor them, there are 8000 files, in total 100 MB. And they keep changing, since last week I've made 2 releases, that'd have added 300 MB to .git folders already. And Brazil is coming, with 7 MB if data, etc. all countries will add up. If this project aims to work on county levels, geospatial data will be large.

If the only problem was the number of files, I could put them in one GeoJSON, but that'd mean slower processing. Still, the problem here is size, and especially handling updated versions, they definitely cannot be in .git

git submodule could work, but then it'd include all 3 quality levels, and also the whole changing .git folder would grow.

Anyway, we just need to fix that one error and it should be good. If we cannot fix it in extract-zip, we just need to use a different unzip module.

hyperknot:
BTW, thanks to @praging the unknown error is fixed. I just needed to fix maxLength in rations.json, I only fixed it in locations.json.

hyperknot:
OK, I have some super great news! yarn supports loading from remote tgz, so this problem is pretty much solved. I'll remove the download zip functions and use the yarn installed one.

It only adds 15 MB gzipped, so I believe it should be ok with all the lambda stuff.

praging:
`coronadatascraper/src/shared/lib/fetch/utils.js
   1:1   error  'extract-zip' should be listed in the project's dependencies. Run 'npm i -S extract-zip' to add it  import/no-extraneous-dependencies
   1:8   error  'extract' is defined but never used                                                                 no-unused-vars
  10:1   error  `adm-zip` import should occur before import of `../fs.js`                                           import/order
  10:20  error  Unable to resolve path to module 'adm-zip'                                                          import/no-unresolved`

hyperknot:
I'm in the middle of a refactor right now, migrating to yarn.

praging:
ah kk

hyperknot:
OK, good news, I've pushed a version which uses yarn add url.tgz and it works wonderfully. I've removed all the dynamic loading code.

Now only two things are left:
- Fix the AUS / _shared folder from being included in location globbing.
- Add some tests for country level functions.

praging:
Do you have a way to pass multiple counties? See https://github.com/covidatlas/coronadatascraper/pull/575.

hyperknot:
Good catch. Right now I think it'd be of great help if we'd extend using ids to 95% of our locations. These special ones could wait.

We can definitely fix the `geography.generateMultiCountyFeature` to work from country-level features, if presented with a code though.

ryanblock:
This kind of dataset sounds like it belongs in a database and/or API - how frequently is it changing?

hyperknot:
@qgolsteyn fixed the scraper handling of _shared folder in 4b632a52, so that is now done.
Literally the only thing missing is adding tests.

This won't update that often, normally only once we add a new region, like Brazil counties now.
I'm confident that this yarn .tgz solution loading a single 15 MB file is the best one for our usecase.

hyperknot:
Went through this with @lazd to make sure everything works. I also added `-c` `--country` CLI option, so we could run:

```
yarn timeseries -d 2020-03-30 -e 2020-03-30 -c AUS
```

So one tiny county is using fips code now in USA, `src/shared/scrapers/USA/CA/shasta-county.js`. It works correctly on the map (and displays in higher resolution then the rest):
image


Also all of AUS is converted to iso codes, so multiple levels can run. It looks right on the map:
```
yarn timeseries -d 2020-03-30 -e 2020-03-30 -c AUS
```

image


Once we finish converting everything, the data will look like this (but it can always be decided afterwards):

```
  {
    "county": "Shasta County",
    "state": "California",
    "country": "United States",
    "maintainers": [
      {
        "name": "John Bencina",
        "github": "jbencina",
        "country": "USA",
        "state": "CA",
        "flag": ":us:"
      }
    ],
    "url": "https://www.co.shasta.ca.us/covid-19/overview",
    "cases": 7,
    "rating": 0.27450980392156865,
    "coordinates": [
      -122.1940585,
      40.735231
    ],
    "tz": [
      "America/Los_Angeles"
    ],
    "featureId": "fips:06089",
    "countryId": "iso1:US",
    "stateId": "iso2:US-CA",
    "countyId": "fips:06089",
    "population": 180080
  }
```

hyperknot:
Test is broken for an unrelated error.

src/events/crawler/scrape-data/normalize-locations.js
```diff
@@ -11,6 +22,38 @@ const normalizeLocations = args => {
...
+                aggregatedCounty.push(countryLevelId);
+              }
+              else {
+                fipsFound = false;
+                log.error('❌ Failed to find FIPS code for subset of combined region %s, %s', aggregatedLocation.county, aggregatedLocation.state);
+              }
+            }
+            if (fipsFound) {
+              // I have no idea if this is useful at all, but it sort of matches what we discussed
+              location.countyId = aggregatedCounty.join('+');
```
hyperknot, line 57: Yes, super useful!

hyperknot: Oh, no we have to set it to the county "name". It'll process it to id at the last step.

src/events/crawler/scrape-data/normalize-locations.js
```diff
@@ -11,6 +22,38 @@ const normalizeLocations = args => {
...
+            }
+          }
+          else {
+            let fipsFound = false;
+            let countryLevelId = findCountryLevelID(location);
+            if (countryLevelId) {
```
hyperknot, line 62: Also this is not saved anywhere.

src/shared/lib/geography/country-levels.js
```diff
@@ -38,6 +38,16 @@ const getLevelData = async level => {
 };
 
 export const getLocationData = async id => {
+  // Split and sum populations for aggregated features (i.e. tricounties)
+  if (id.indexOf('+') !== -1) {
+    const parts = id.split('+');
+    let data = await Promise.all(parts.map(getLocationData));
+    data = data.reduce((a, d) => {
+      a.population += d.population;
+      return a;
```
hyperknot: We could also do something with the name here, right? Like joining with &, like `Abc & Def`.

src/shared/lib/geography/index.js
```diff
@@ -104,7 +104,8 @@ export const isCity = function(location) {
 };
 
 export const getSmallestLocationStr = function(location) {
-  const smallestStr = location.city || location.county || location.state || location.country;
+  const county = !location.countyId ? location.county : false;
```
hyperknot, line 107: I guess that's the aggregate case, right? Why don't we overwrite the county string, like in other places?

lazd: Yes, this is the `TriCounties, Utah` case. It was causing issues to leave the IDs in `county`, since the name cannot be resolved, and we already have made a human readable name

src/shared/lib/geography/country-levels.js
lazd: Agreed, we should. Right now, it's done manually inside of the scraper.

src/events/crawler/scrape-data/normalize-locations.js
lazd: I think you're right, especially with the other comment on combining names.

PR #529, opened 2020-03-30
ARCGIS for MA

aed3:


cmcjacob:
Question, any reason you went with json here instead of pulling the csv? It's the same data so doesn't really matter - just curious

aed3:
> Question, any reason you went with json here instead of pulling the csv? It's the same data so doesn't really matter - just curious

It's easier for me to find the JSON link ¯\_(ツ)_/¯.  I can just copy the correct query from the map and it gives me the data.

src/shared/scrapers/USA/MA/index.js
```diff
@@ -114,6 +119,56 @@ const scraper = {
...
+      let onlySumDeaths = true;
+      let sumDeaths = 0;
+
+      data.features.forEach(item => {
+        const cases = item.attributes.CASES;
+        const county = geography.addCounty(
+          item.attributes.County.charAt(0) + item.attributes.County.slice(1).toLowerCase()
+        );
+
+        if (datetime.scrapeDateIsBefore(item.attributes.EditDate)) {
```
chunder: Is this logic right? Don't you want a NOT in front here...

if ( ! datetime.scrapeDateIsBefore(item.attributes.EditDate)) { throw ... }

aed3: It is. It confused me for a while too because I think the function naming is backwards, but I manually tested it with different dates and it worked

PR #530, opened 2020-03-30
fix: realign table values for data accuracy, create location jsons, update statePopulation json

ArnavD74:
-Data source changed their table slightly, so the column order scraped was changed accordingly.
-Added jsons for latitude and longitude to be used for geo
-Updated state population json entry order to match with data source

-Scraper still can not find these location values despite pushing the lat and long for each location through the jsons: 
❌ Could not find location Ladakh, IND (new territory)                                                 
❌Could not find location Telengana, IND (spelling difference)


ddaash:
if you can check on  annotation failure on CI build.

ArnavD74:
> if you can check on  annotation failure on CI build. 

I don't believe that the issue is related to changing the Scraper

ArnavD74:
The entire sources data format changed again so I will have to take a look.

PR #532, opened 2020-03-30
ARCCGIS for CT + CO Deaths

aed3:


src/shared/scrapers/USA/CT/index.js
```diff
@@ -140,6 +140,33 @@ const scraper = {
...
+      const counties = [];
+
+      data.features.forEach(item => {
+        const cases = item.attributes.ConfirmedCases;
+        const deaths = item.attributes.Deaths;
+        const county = geography.addCounty(item.attributes.COUNTY);
+
+        if (datetime.scrapeDateIsBefore(item.attributes.DateLastUpdated)) {
+          throw new Error(`Data only available until ${new Date(item.attributes.CV_Updated).toLocaleString()}`);
+        }
```
chunder: You reference item.attributes.DateLastUpdated in the if statement, but item.attributes.CV_Updated in the throw text. Intentional?

aed3: NOPE! That's a whole mistake. I'll fix it right now!

src/shared/scrapers/USA/CO/index.js
```diff
@@ -141,6 +141,7 @@ const scraper = {
       }
       const stateData = transform.sumData(counties);
       stateData.population = parse.number(data[0].State_Population);
+      stateData.deaths = parse.number(data[0].State_Deaths);
```
chunder: Does this work on older cached versions? Has the State_Deaths field been these since we started scraping it?

aed3: Yes, the data was always there
Also realized I was double wrong about `datetime.scrapeDateIsBefore()`. It should be `datetime.scrapeDateIsAfter()`

lazd:
Ok tested time series for each of these:

```
yarn timeseries -l "CO, USA" -d 2020-3-12 -e 2020-3-29
```

and 

```
yarn timeseries -l "CT, USA" -d 2020-3-14 -e 2020-3-29
```

Both work great!

PR #534, opened 2020-03-30
Added US Virgin Islands.

jacobmcgowan:


lazd:
Looks like this duplicates #511, @ParkerM can you two work together to figure out the best implementation, and figure out how to get it to match the right features and population on the map? Thanks!

src/shared/scrapers/VIR/index.js
```diff
@@ -0,0 +1,67 @@
+import * as fetch from '../../lib/fetch/index.js';
+import * as parse from '../../lib/parse.js';
+
+const scraper = {
+  country: 'VIR',
+  type: 'table',
+  url: 'https://doh.vi.gov/covid19usvi',
+  sources: [
+    {
+      url: 'https://doh.vi.gov',
```
chunder: I would link this url right to the covid page: https://doh.vi.gov/covid19usvi

src/shared/scrapers/VIR/index.js
```diff
@@ -0,0 +1,67 @@
...
+  ],
+  async scraper() {
+    const results = {
+      cases: 0,
+      tested: 0,
+      deaths: 0,
+      active: 0
+    };
+    const $ = await fetch.page(this.url);
+    const $widget = $('div.widget:has(div.title-box:has(h1:contains(COVID-19 Cases)))');
```
chunder: Can you check to be sure you find the widget and throw an error if not? They may change the page.

src/shared/scrapers/VIR/index.js
```diff
@@ -0,0 +1,67 @@
...
+          results.deaths = parse.number(value);
+          break;
+        case 'dead':
+          results.deaths = parse.number(value);
+          break;
+        default:
+          break;
+      }
+    });
+
```
chunder: Or you could throw an error down here if cases, negative, or tested are 0. We know they must be >0 now. If they are 0 now, it's because the page changed and your scraper isn't working with the new page format.

ParkerM:
> Looks like this duplicates #511, @ParkerM can you two work together to figure out the best implementation, and figure out how to get it to match the right features and population on the map? Thanks!

I'm cool with closing #511 in favor of this one, it seems clearer and better thought out. I'm also pretty unfamiliar with the domain and it looks like @jacobmcgowan has a better grasp on how the data is actually consumed 👍.

src/shared/scrapers/VIR/index.js
jacobmcgowan: Added error checking. Do you think the wording for the error message is fine or should I word it differently?

lazd:
Works great, thanks!

PR #535, opened 2020-03-30
Added active testing for Puerto Rico.

jacobmcgowan:


lazd:
This is confusing: we take `active` to mean non-recovered or deceased cases. There is no way `active` can be higher than `cases` if we accept that definition. What is the `active` field coming from Puerto Rico supposed to mean?

jacobmcgowan:
There have been some confusion on my end. I assumed active to mean actively being tested. I will abandon this PR.

PR #536, opened 2020-03-30
fix: MA & CT Date error check

aed3:
It does work as it currently is, but if for some reason the scraper date is one day behind the data's date, it won't work properly


lazd:
Hmm, I wonder what it is parsing those dates as? Maybe it's getting the wrong date when it parses... Can you figure out why it was failing before?

aed3:
> Hmm, I wonder what it is parsing those dates as? Maybe it's getting the wrong date when it parses... Can you figure out why it was failing before?

Yeah, I just did. It's telling me the data is from `3/29/2020, 4:34:12 PM` and the scraper date is `3/30/2020, 11:02:22 AM`, however if I put the milliseconds the data is giving me into Chrome, it give me `3/30/2020 04:01:42 AM`. The timezone of the scraper and the dataset don't match which is causing the problem. I'm not sure how to get around that though.

lazd:
@aed3 it should be resetting the hours for the comparison inside of `datetime.js`... That should handle any of the screwups.

```
  localA.setHours(0, 0, 0, 0);
```

aed3:
Why is there the minus 7 in the below code from `datetime.js`? Removing it fixes the problem, but will probably cause others.

```
export const getDate = function() {
  const date = new Date();
  const utcDate = new Date(date.toUTCString());
  utcDate.setHours(utcDate.getHours() - 7);
  return new Date(utcDate);
}
```

lazd:
@aed3 to get things in PST

Issue #179, opened 2020-03-20
Add scrapeDate column to data


lazd:
This has been requested a few times, and it seems to be required now that we might release data with stale timeseries data...

feliperoberto:
I saw https://github.com/lazd/coronadatascraper/commit/efc561e6d23b054f33b41ce8c67ac904983a1628 , what still should be done?

jzohrab:
Spec notes from slack conversation with lazd:

We should add a "scrapeDate" field (name TBD) to `dist.json` giving the date of the cache file that was used for the scrape.  e.g.

```
{
    "state": "PA",
    "country": "USA",
    "aggregate": "county",
    ...
    "population": 102811,
    "scrapeDate": blah  <<<<<<<<<<<<<<
  },
``` 

slack convo notes:

```
Larry Davis: it should be the datetime of the cache file being used
jz: right.  of the folder it’s in?
Larry Davis: oh the file itself, really… but even then that modified date is probably wrong
jz: my cache has folders 2020-3-27 etc.  Is that the same in prod?
Larry D: should be
```

hyperknot:
I don't know where to put it in a timeseries CSV, probably nowhere. But in a timeseries JSON it'd be a nice bonus.

And it really should be a scrape time in local timezone, not scrape date.

To make this project really high quality, it should run every hour, and keep the latest time for each day. So for example, it'd scrape at 20:50, 21:50, 22:50, 23:50 local time and then keep the number of the 23:50 one. scrape_time would be then saved as `2020-03-30T23:50:00+02:00`

@HerbCaudill

hyperknot:
@ryanblock

hyperknot:
@camjc

hyperknot:
But I totally agree about logic behind it, it should be the datetime of the cache file used to compute the values.

chunder:
It should *not* be in a local timezone. Recording of the time of events should be in UTC. You can translate that to any timezone for display purposes if desired.

chunder:
I am also interested in adding an `updatedAt` field to the records returned by scrapers. Many data sources explicitly note `Last updated at XXXX`.

@lazd and I have been talking about having scrapers throw errors if they're scraping data which hasn't been updated in more than 1 day. That likely means they are scraping a source which is no longer being updated. Agencies often switch to a new web page or csv file or somesuch and just leave the old page/file hanging around with stale data in it.

hyperknot:
[...]

PR #537, opened 2020-03-30
Switch Canada to updated csv url; throw an error if it gets stale

chunder:


src/shared/scrapers/CAN/index.js
```diff
@@ -45,27 +46,23 @@ const scraper = {
...
+      const scrapeDate = process.env.SCRAPE_DATE ? new Date(`${process.env.SCRAPE_DATE} 12:00:00`) : datetime.getDate();
+      const scrapeDateString = datetime.getDDMMYYYY(scrapeDate);
       const lastDateParts = data[data.length - 1].date.split('-');
       const lastDateInTimeseries = new Date(`${lastDateParts[2]}-${lastDateParts[1]}-${lastDateParts[0]} 12:00:00`);
       const firstDateParts = data[data.length - 1].date.split('-');
       const firstDateInTimeseries = new Date(`${firstDateParts[2]}-${firstDateParts[1]}-${firstDateParts[0]}12:00:00`);
 
       if (scrapeDate > lastDateInTimeseries) {
-        console.error(
+        throw new Error(
```
lazd: So we're currently taking the old data anyway, not throwing. Any reason for the change?

chunder: I'm tired of scrapers pulling old data for days with no one noticing.

I think this code will work and return results for all the dates which are in the csv. But it will throw when you ask it for a date later than what is contained in the csv.

chunder: ℹ️  Total counts (tracked cases, may contain duplicates):
   - 10746 cases
   - 12484 cases
   - 12608 cases
  ❌ Error processing CAN:  Error:   🚨 timeseries for CAN: SCRAPE_DATE 31-03-2020 is newer than last sample time 2020-3-30.
   - 0 cases

lazd: Ok, I hear that. Unfortunately it will mean that Canada will disappear from the dataset for a day if they're slow to update, which means they won't show up on the map... Maybe we should just be more aggressive about the warning? I know the log level work @ryanblock is doing will make it so we only see errors, so that will help for human eyes, and if we implement #179 and have a QA check on whether the date has matches the current date, we can throw and let the status bot know... 

@qgolsteyn @ryanblock @camjc thoughts?

chunder:
Why does it disappear?

The timeseries command returns good data for all the valid dates?

src/shared/scrapers/CAN/index.js
qgolsteyn: It would disappear from `data.json/csv` if it is not updated. Ideally, I think we should still include Canada with some sort of flag to indicate the data is stale

lazd: We will have it throw when it becomes 2 days stale. For now, I am reverting to always accepting and merging this tonight, it's late and we can fix it tomorrow.

PR #539, opened 2020-03-30
Fixed MS to handle new page html

chunder:


lazd:
Nice work, thank you so much.

PR #540, opened 2020-03-30
ARCGIS for MN

aed3:


lazd:
Looks good and works fine, wondering why we're using the JSON instead of the CSV via `fetch.getArcGISCSVURL`? @chunder insights?

lazd:
I don't prefer one to the other, I just want it to work if something gets updated, and I'm not sure which is more stable.

chunder:
I don’t have any reason to prefer one way over another. This code checks to be sure it’s getting freshly updated data, which will at least shout at us if it gets stale.

PR #547, opened 2020-03-30
update: AZ uses arcgis instead of tableau

cmcjacob:


chunder:
Does the CSV contain any timestamps showing when the data was last updated? I'm trying to make sure our crawlers can detect when they are scraping stale data.

@lazd and I were talking about having scrapers throw an error if the data they find was last updated more than 1 day ago.

cmcjacob:
Unfortunately this state is not adding update timestamps to [their dashboard](http://azdema.maps.arcgis.com/apps/opsdashboard/index.html#/331b8b9d7a764017a21815f6734e11ff) yet.  Only a handful are, currently.  Could stale data be detected from unchanging data in cases we don't have update timestamps?

src/shared/scrapers/USA/AZ/index.js
```diff
@@ -18,35 +18,53 @@ const scraper = {
...
 
-      counties.push({
-        county,
-        cases
-      });
-    }
+      counties.push(transform.sumData(counties));
+      return counties;
+    },
+    '2020-3-20': async function() {
```
lazd: did you mean `2020-3-30`?

src/shared/scrapers/USA/AZ/index.js
```diff
@@ -18,35 +18,53 @@ const scraper = {
...
+          // unfortunately even arcgis isnt reporting any death data
+          county: county.NAME,
+          cases: parse.number(county.Number_Confirmed || 0)
+        });
+      }
 
-    counties.push(transform.sumData(counties));
-    return counties;
+      counties.push(transform.sumData(counties));
+      console.log(counties);
```
lazd:
```suggestion
```

PR #555, opened 2020-03-31
Add scrapeDate to dist.json -- PR for discussion only

jzohrab:
An attempt at #179.  I attempted to explain this in Slack, but it appears that my meaning was lost, so here are the comments again.  Hopefully this will clarify the argument, so someone can point out the mistakes in my thoughts.

---

All scrapers seem to be using the libraries in src/shared/lib/fetch/index.js (at least, I hope they are, haven’t triple-checked).  The exported methods are page,  json, csv, tsv, pdf, and headless (omitting the getArcGIS* methods, which don’t mention date).

They all have the same declarations: `const <methodname> = async (url, date, options)`, and for each date = “date the date associated with this resource, or false if a timeseries data”
Checking the calls, none of the scrapers ever pass the date, or if they do, it’s “false”. (I can show the grep search results for each if anyone needs it) 

All of these methods (except for headless) call `get.js`’s
`export const get = async (url, type, date = process.env.SCRAPE_DATE || datetime.getYYYYMD(), options = {})`
headless has `date = process.env.SCRAPE_DATE || datetime.getYYYYMD()` for its args.

In other words, date is always going to be `process.env.SCRAPE_DATE || datetime.getYYYYMD()`

Since every call uses the `process.env.SCRAPE_DATE || datetime.getYYYYMD()` for its cache lookup, it seems that we could just add that to as the `scrapeDate`

### DIFF
--- a/src/events/processor/write-data/index.js
+++ b/src/events/processor/write-data/index.js
@@ -2,6 +2,7 @@ import path from 'path';
 import * as fs from '../../../shared/lib/fs.js';
 import * as stringify from './stringify.js';
 import reporter from '../../../shared/lib/error-reporter.js';
+import * as datetime from '../../../shared/lib/datetime.js';
 
 const writeData = async ({ locations, featureCollection, report, options, sourceRatings }) => {
   let suffix = '';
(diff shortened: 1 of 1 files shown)


lazd:
@jzohrab see https://github.com/covidatlas/coronadatascraper/commit/5dda6a5ce58c1bcd04c5d61e1b486a6d42fed984

The schema has been updated to support both `collectedDate` (the date we pulled the file) and `publishedDate` (passed down by scrapers).

All that's left is somehow getting the `collectedDate` from the caching mechanism down into the data pipeline. I bet @ryanblock can pull it off.

jzohrab:
Thanks @lazd .  I think this PR code is completely incorrect.  You?

jzohrab:
Yep, it's bad, closing.

PR #560, opened 2020-03-31
COVIDAtlas.com frontend work

ryanblock:


lazd:
@NateBaldwinDesign @JacobDFrank this is the branch where the frontend work can happen.

PR #563, opened 2020-03-31
Skip the TOTALS row that was just added for MA, USA

chunder:


chunder:
@qgolsteyn @aed3 Can you review?

src/shared/scrapers/USA/MA/index.js
```diff
@@ -132,6 +132,10 @@ const scraper = {
       let sumDeaths = 0;
 
       data.features.forEach(item => {
+        if (item.attributes.County === 'TOTALS') {
+          return;
```
aed3, line 136: If we're including hospitalization too now, then you can get that from the totals row and then return. But that'd be a future change.

src/shared/scrapers/USA/MA/index.js
```diff
@@ -132,6 +132,10 @@ const scraper = {
       let sumDeaths = 0;
 
       data.features.forEach(item => {
+        if (item.attributes.County === 'TOTALS') {
```
qgolsteyn, line 135: To avoid a potential regression, would be possible to make this check case insensitive?

PR #564, opened 2020-03-31
Fixed new header sturcture in ND

chunder:


chunder:
@qgolsteyn @aed3 Can you review?

PR #571, opened 2020-04-01
Added South Korea scraper.

jacobmcgowan:


jacobmcgowan:
Found an issue with the reported deaths. Will fix and resubmit.

src/shared/scrapers/KOR/index.js
```diff
@@ -0,0 +1,88 @@
...
+   * @returns {string} The common name of the province or the city name if a
+   * special city was given.
+   */
+  _parseProvince(officialEnglishName) {
+    return this._provinceNames[officialEnglishName] || officialEnglishName;
+  },
+  async scraper() {
+    const states = [];
+    const $ = await fetch.page(this.url);
+    const $table = $('table.num');
```
chunder, line 68: Can you test to be sure you found the table? 

`if ($table.length === 0) { throw new Error('Table not found'); }`

chunder: Can you also check each column header against your `fieldPositions` to be sure you throw an error if they change the table layout?

lazd:
@qgolsteyn @chunder either of your CR this?

Issue #363, opened 2020-03-26
Irregular state/county/city names in timeseries-byLocation.json


NickSto:
I was looking through `timeseries-byLocation.json` and found some odd-looking place names. I thought I'd document them here, since they seem likely to be errors.

Weird county names:
- `(unassigned)` in Arkansas and Washington.
- `County` in Pennsylvania.
- `Not Reported County` in Michigan
- `Pending County` in Tennessee.
- `Tri County` in Utah
  - I checked, and that doesn't seem to be a real county.
  - It's probably a common name for a region.

`Bear River County` is listed as a city in Utah (no county name given).
- There is a "Bear River City" in Utah, so maybe that's what it is?
  - It's in "Box Elder County"

There's a `state` named `Total` in Canada. There's no Canadian entry without a `state`, so I'm assuming that serves as the Canada-wide entry. Still, it'll flummox any parser expecting to recognize a country-wide entry by its lack of a `state`.

There may be more issues; I didn't have time to look carefully. But I scanned through the whole list.

mdibenigno:
I encountered this as well. The USA has country total rows (no state, no county), but China does not have country total rows and only at the state level... My parser is looking pretty bad with all of these data inconsistencies.

lazd:
@mdibenigno yeah, we didn't roll up China's data. Care to send a PR to do so?

mdibenigno:
@lazd ah got you... also great to know it is just China... was about to start doing a giant check for other countries... thanks for sharing the link to the other issue!

lazd:
@mdibenigno it was AUS too, but that's rolled up back to the start.

tautme:
It seems Utah is reporting Jurisdictions, which include multiple counties.
[https://coronavirus-dashboard.utah.gov/](https://coronavirus-dashboard.utah.gov/)

NickSto:
Are those even actual legal entities? Or is it just how the Department of Health decided to divide up reporting?

I guess either way, you could just decide that's what will serve as for the `county` field for Utah and report verbatim.

tautme:
It seems the population is too low to have individual administration. I have been thinking about it as Country, then subCountry, subsubCountry, and so on....

praging:
I couldn't find the CAN issue. Could you verify if it has been fixed?

mark-otaris:
[...]

PR #575, opened 2020-04-01
Fix Issue 363: Changes to UT & MI, USA

praging:
Fixes `Not Reported County` in Michigan as well as `Tri County Health District` and `Bear River Health District` in Utah.

Closes https://github.com/covidatlas/coronadatascraper/issues/363.


lazd:
NICE. I never knew what Bear River was. Thanks for doing the work!

PR #578, opened 2020-04-01
Add `timeseries` to ratings schema

HerbCaudill:
Just noticed that ESP was showing up as not a timeseries, then noticed that **all** sources where showing up as not a time series, whether they are or not. 

image

Tracked it down to the ratings schema, which was missing the `timeseries` property.


lazd:
SHIT. nice catch.

PR #579, opened 2020-04-01
ESP metadata: add type, fix sources

HerbCaudill:
On the ESP scraper definition:
- I'd left out the `type` property, which is csv
- I'd misunderstood the `sources` property; I've renamed that to `_endpoints` (for internal use by the scraper) and populated the original sources correctly (Spain Min. of Health & Dept. of National Security).


jzohrab:
LGTM

lazd:
Nice one thanks!

PR #582, opened 2020-04-01
feat: move running crawler for timeseries to its own process

qgolsteyn:
`yarn timeseries` has been running out of memory on many occasions. To prevent this from occurring, we can run the crawler in its own process. This PR adds a new function `runCrawler` which is called from `timeseries/index.js` and takes care of creating a child process and retrieving data from this process when crawling is done.

### DIFF
--- a/src/shared/cli/cli-args.js
+++ b/src/shared/cli/cli-args.js
@@ -31,6 +31,18 @@ const { argv } = yargs
     description: 'Suppress logs',
     type: 'boolean'
   })
+  .option('findFeatures', {
+    description: 'Include feature information in output data',
+    type: 'boolean'
+  })
+  .options('findPopulations', {
+    description: 'Include population information in output data',
+    type: 'boolean'
+  })
+  .options('writeData', {
+    description: 'Write to dist folder',
+    type: 'boolean'
+  })
   .help()
   .alias('help', 'h');
 
--- a/src/shared/timeseries/index.js
+++ b/src/shared/timeseries/index.js
@@ -2,14 +2,14 @@ const imports = require('esm')(module);
 
 const path = require('path');
 
-const generate = imports('./index.js').default;
-const argv = imports('./cli/cli-args.js').default;
-const fs = imports('./lib/fs.js');
-const transform = imports('./lib/transform.js');
-const geography = imports('./lib/geography/index.js');
-const datetime = imports('./lib/datetime/index.js').default;
+const argv = imports('../cli/cli-args.js').default;
+const fs = imports('../lib/fs.js');
+const transform = imports('../lib/transform.js');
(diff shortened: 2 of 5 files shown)


chunder:
Looks like a nice clean refactor to me. Caveat: I'm not familiar with the js conventions for calling child processes.

src/shared/cli/cli-args.js
```diff
@@ -31,6 +31,18 @@ const { argv } = yargs
...
+    type: 'boolean'
+  })
+  .options('findPopulations', {
+    description: 'Include population information in output data',
+    type: 'boolean'
+  })
+  .options('writeData', {
+    description: 'Write to dist folder',
+    type: 'boolean'
+  })
```
lazd, line 45: Nice!

src/shared/timeseries/runCrawler.js
```diff
@@ -0,0 +1,34 @@
...
+ * This will result in a slower execution time due to the child_process overheads,
+ * but it will ensure that timeseries never runs out of memory.
+ */
+export default options =>
+  new Promise((resolve, reject) => {
+    const child = childProcess.fork(
+      path.join('src', 'shared', 'timeseries', 'worker.js'),
+      [
+        '-d',
+        options.date,
```
lazd, line 16: Can you also pass `-e` and `-o` and `-q` and really all of them?

src/shared/timeseries/index.js
```diff
@@ -190,7 +190,8 @@ async function generateTimeseries(options = {}) {
   const lastDate = dates[dates.length - 1];
   let featureCollection;
   for (const date of dates) {
-    const data = await generate(date === today ? undefined : date, {
+    const data = await runCrawler({
+      date: date === today ? undefined : date,
```
lazd, line 194: Much cleaner

src/shared/timeseries/index.js
```diff
@@ -190,7 +190,8 @@ async function generateTimeseries(options = {}) {
   const lastDate = dates[dates.length - 1];
   let featureCollection;
   for (const date of dates) {
-    const data = await generate(date === today ? undefined : date, {
+    const data = await runCrawler({
```
lazd, line 193: For a second, I thought it would be fun push all of the promises to an array and do an `await Promise.all(promisesArrray)`, but then I realized we'd probably run out of memory even faster :)

qgolsteyn: Yes, that would probably break things haha. Let's start with keeping this process sequential.

src/shared/timeseries/runCrawler.js
qgolsteyn: Done!

PR #591, opened 2020-04-02
revert: "Add county level scraper & GeoJSON for Brazil"

qgolsteyn:
Reverts covidatlas/coronadatascraper#476 to allow for our new location ID to be merged


lazd:
I think we're good, location ID is working with the latest master

lazd:
Sounds like @hyperknot wants this revered still.

hyperknot:
Yes, I'm working with @thiagoghisi on an ID based PR.

PR #592, opened 2020-04-02
[WIP] Rebuilding caching to be timezone aware and use 8601Z

ryanblock:


src/shared/lib/datetime/iso/cast.js
```diff
@@ -0,0 +1,13 @@
+import { ZoneId, ZonedDateTime } from '@js-joda/core';
+import '@js-joda/timezone/dist/js-joda-timezone-10-year-range';
+
+/**
+ * Returns locale-cast YYYY-MM-DD from a UTC-cast 8601Z
+ * @param {*} tz IANA timezone string, e.g. `America/Los_Angeles`
+ * @param {*} utc ISO8601Z string, e.g. `2020-04-02T01:23:45:678Z`
```
hyperknot, line 7: This is a super minor thing, but what is `:678Z` at the end? 

In my understanding a standard ISO datetime is: `2020-04-02T12:18:17+02:00`

There could be fractional seconds, which I think we shouldn't be using, so `2020-04-02T12:18:17.912436+02:00` is also valid but probably wrong for our usecase.

hyperknot: BTW, this is also cutting it to data if we have `.split('T')[0]`, so it's actually returning something different.

ryanblock: 8601 is not just a single format, it covers a wide variety of implementations; `2020-04-02T12:18:17+02:00` is one such implementation. I typoed a colon instead of a period there, but see: `new Date().toISOString() // '2020-04-02T13:47:23.243Z'`

hyperknot: OK, clear now.

ryanblock:
Ok, I'm feeling p decent about this now.

To make this fully ready, at the time of cache retrieval I need the IANA timezone for the scraper in question. cc @hyperknot 

We could (in theory) merge without it as is, but I really need that timezone support as a fast-follow – and we've got to have it exist for every single scraper, otherwise we leave every non-timeseries data point open to off by one errors.

ryanblock:
Following up: I'm now seeing how we're relying on scraper logic to define caching, which can create a highly coupled and brittle caching path. Will be proposing some related changes shortly.

jzohrab:
Hi @ryanblock - perhaps this PR should be closed.  Thoughts?

ryanblock:
Yeah let’s kill it, I extracted all my work from there that was needed for
Li.

jzohrab:
Closing, not relevant now for this project.  Thanks!

PR #595, opened 2020-04-02
space option in writeJSON

hyperknot:
## Summary

1. Pass "space" option to JSON.stringify in fs.writeJSON
2. Generate features.json with space: 0

Makes the file size from 75 MB to 25 MB.


jzohrab:
LGTM, thanks @hyperknot !

PR #598, opened 2020-04-02
Tableau data for KS

aed3:
## Summary

Allows tableau data for Kansas counties to be parsed

## Changes

Grabs data from the PDF version of the table containing cases, sorts the characters of the scraped results so they're in the same order they appear on the document, adds the characters together to get county names and case numbers, and gets the total death count for the state from a different sheet within the same tableau dashboard.

## Additional notes

Where did this template come from? I like it!


lazd:
Merged as part of #613

PR #603, opened 2020-04-02
add Recovered for AUS-SA

camjc:
## Summary

South Australia added a table row, which throws an Error that we noticed in the scraper report. Now we have a recovered count, just a tested count still to go if they ever add it.
This is a backwards compatible change.

### DIFF
--- a/src/shared/scrapers/AUS/SA/index.js
+++ b/src/shared/scrapers/AUS/SA/index.js
@@ -4,7 +4,12 @@ import * as parse from '../../../lib/parse.js';
 import getKey from '../_shared/get-key.js';
 import maintainers from '../../../lib/maintainers.js';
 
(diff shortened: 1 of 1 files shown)


src/shared/scrapers/AUS/SA/index.js
```diff
@@ -4,7 +4,12 @@ import * as parse from '../../../lib/parse.js';
 import getKey from '../_shared/get-key.js';
 import maintainers from '../../../lib/maintainers.js';
 
-const labelFragmentsByKey = [{ cases: 'confirmed case' }, { hospitalized: 'icu' }, { deaths: 'deaths' }];
+const labelFragmentsByKey = [
+  { cases: 'confirmed case' },
+  { deaths: 'deaths' },
+  { hospitalized: 'icu' },
+  { recovered: 'cases cleared' }
```
camjc, line 11: New item

PR #606, opened 2020-04-02
fix: Utah, USA now fleshed out

praging:
added Southeast, Southwest, and Central Health Districts.


lazd:
Oh damn, we were completely misrepresenting Utah. GOOD CATCH!

PR #607, opened 2020-04-03
update test for AK

hyperknot:
@lazd the tests are updated for AK. Does it look OK for you? 

I don't understand why did the original test have a very buggy looking line `{ "cases": 42 }` in there, but now it looks much better.


hyperknot:
Oh, that 42 is the `transform.sumData(counties)`, that's a weird pattern, but at least I know it's good.

PR #608, opened 2020-04-03
fixes #552: LA, USA

### LINKED ISSUE #552: LA, USA: loc / pop failure
location and population gets begin failing 2020-3-18 in backwards chronological order. seems as though "Parish" is not being added to the raw name. unsure as to how these older dates are being handled in the scraper.

praging:


praging:
this commit fixes VA rather than LA. renamed and resubmitted as #609.

PR #609, opened 2020-04-03
fixes VA, USA city loc / pop issue

praging:


lazd:
Clean. Though even if the capitalization is wrong, it should have still worked. @hyperknot let's account for this in the future.

hyperknot:
@lazd I don't understand why is this needed? Isn't our fips-code finding case-insensitive already?

PR #611, opened 2020-04-03
Notes on installing on a fresh machine

shaperilio:
Quick note on getting started since on a fresh machine you may end up with the wrong version of Node. (The fact that mine is Catalina may be irrelevant)

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -28,6 +28,7 @@ git submodule update
 ```
 yarn install
 ```
+Note: on a Mac (Catalina) simply running `brew install yarn` may give you an incompatible version of `node`. If that's the case, [install the n package](https://www.npmjs.com/package/n) and run, e.g. `n lts`.
 
 ##### 3. Run the scraper
 


jzohrab:
LGTM, docs change only, no code.

jzohrab:
> Is yarn or brew giving you node? Or It’s whatever comes with your OS?

It can be whatever's on your machine, or maybe you have another proj w/ node.  n's a decent solution for ppl who are new to JS.

shaperilio:
> Is yarn or brew giving you node? Or It’s whatever comes with your OS?

This is a brand new machine, so when I installed `yarn` with `brew` it installed the latest `node` and then I had to downgrade to the latest LTS.

hyperknot:
Here is my trusty macOS node howto, developed over the years. Yarn should definitely not be installed from brew, use the curl method to do it.

Should we include it?

```
n-uninstall -y

rm -rf /usr/local/lib/node_modules
rm -rf ~/.npm 
rm -rf ~/.node*
rm -rf ~/Library/Caches/Yarn
rm -rf ~/n
rm -rf ~/.config/yarn
rm -rf ~/.config/configstore
rm -rf ~/.yarnrc
rm -rf ~/.standard-cache

brew uninstall --force node@0.12 node@4 node@6 node yarn
brew cleanup


#
#
# reopen terminal window
#
#


# INSTALL

#curl -L https://git.io/n-install | bash

curl -L https://git.io/n-install | bash -s -- -y lts


#
#
# reopen terminal window
#
#

#n 12
n prune

curl -o- -L https://yarnpkg.com/install.sh | bash
```

hyperknot:
Basically the key is to use `n-install` (or nvm) to select the right node version and then `curl -o- -L https://yarnpkg.com/install.sh | bash` to install yarn.

In my opinion, you should not use homebrew for either node or yarn!

jzohrab:
This one-line readme PR has become something. :-P  @hyperknot , I installed yarn with brew :-) and it worked -- which was good enough for me, as I was new to JS and had no idea what was up.  Your instructions look good but would have scared me off, I think!  I'm always reluctant to curl bash, even though I know it's usually ok.

How about we change this PR to a suggestion on how to handle it for newbies (which I would have liked at the time), and let more experienced people do what works for them?

hyperknot:
@jzohrab basically the only way to install yarn on non-homebrew installed node is to use the curl script. I don't understand why is it any more scarier than any npm install, here at least I know I'm installing the official code from yarnpkg.com, whereas all the random stuff which happens on my computer in a npm/yarn install.

jzohrab:
Thanks for input from all, and thanks @shaperilio for starting this.  I pulled in changes and added some minor changes in https://github.com/covidatlas/coronadatascraper/pull/669, so closing this in favor of that.

PR #619, opened 2020-04-03
Add eslint plugin `unicorn`, mainly to enforce file names as kebab case

camjc:
Add eslint plugin `unicorn`, mainly to enforce file names as kebab case.
Added and auto-fixed other rules that seem appropriate (see .eslintrc.json).

These mainly just make the semantics of operations a little clearer; like checking includes rather than !== -1

Primary impetus for this PR was to enforce lowercase filenames, as this is a now agreed-upon standard, and so linting is easier than picking up at CR time.


.eslintrc.json
```diff
@@ -16,20 +22,51 @@
...
+    "unicorn/prefer-node-remove": "error",
+    "unicorn/prefer-number-properties": "error",
+    "unicorn/prefer-reflect-apply": "error",
+    "unicorn/prefer-replace-all": "error",
+    "unicorn/prefer-set-has": "error",
+    "unicorn/prefer-spread": "error",
+    "unicorn/prefer-text-content": "error",
+    "unicorn/prefer-trim-start-end": "error",
+    "unicorn/prefer-type-error": "error",
+    "unicorn/throw-new-error": "error"
```
camjc, line 70: These are all auto-fixed in this PR, and seem generally good. Happy to remove any we think inappropriate.

camjc: I especially like the one that turns all the `-1` checks into `includes`

camjc:
@jzohrab While we're adding that unicorn eslint plugin I figured we may as well apply some of the auto-fixable rules at the same time (I especially enjoy using `includes` over checking `-1`s). Happy to decouple or just do the filename thing if we don't like the other changes.

camjc:
This does far too much, closing and will do little PRs :)

PR #620, opened 2020-04-03
Fix scraping of JHU 'recovered' data

### LINKED ISSUE #618: Data obtained from JHU missing 'recovered' since 2020-03-24
As the subject goes.

On 2020-03-24 a new format of data was introduced which moved data about 'recovered' to a separate file. The applicable code does not handle this properly.

yoctozepto:
Note the 'US' conditional branch is never hit there because of
the lack of US states in the global data (separate issue).

Closes #618

### DIFF
--- a/src/shared/scrapers/JHU.js
+++ b/src/shared/scrapers/JHU.js
@@ -294,12 +294,11 @@ const scraper = {
             country: 'USA',
             state,
             cases: parse.number(cases[index][date] || 0),
-            deaths: parse.number(deaths[index][date] || 0),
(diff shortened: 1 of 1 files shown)


lazd:
The US branch will get hit for old data, but this fix looks good. Will give it a test and merge

yoctozepto:
@lazd I beg to differ, this is already variant for the new data so it will not get hit.

lazd:
@yoctozepto ah, so there is still a useless branch? Does it seem to be affecting anything?

lazd:
Yup, you're totally right about that, and this fix works great. Thanks!

PR #632, opened 2020-04-03
Add integration test data files only.

jzohrab:
##  Summary

Data files in `tests/integration/scrapers/testcache` only, zero code impact.


lazd:
I like test data

PR #636, opened 2020-04-03
Scraper tests by date

### LINKED ISSUE #450: scrapers.all.test.js doesn't handle cases where a scraper has multiple scrapers defined
`scraper.all.test.js` has the following:

```
      it('returns latest data', async () => {
        const expectedPath = join(testDir, 'expected.json');
        let result = await scraperObj.scraper();
```

For some scrapers, such as USA/AL, `scraperObj.scraper` is an object, not a function: `{ '0': [AsyncFunction: 0], '2020-3-26': [AsyncFunction: 2020-3-26] }`

Currently, the tests are passing in master due to a hack added to `scraper.all.test.js`:

```
      if (scraperObj.state === 'AL' && scraperObj.country === 'USA') {
        // Honestly these linter rules are absurd
        // eslint-disable-next-line
        scraperObj.scraper = scraperObj.scraper[0];
      }
```

This testing idea should support multiple input file formats, just like it supports multiple expected file outputs.

jzohrab:
## Summary

This closes #450, tests for scrapers on multiple dates.

Prior to this change the tests would not respect date updates for scrapers (date versioning).

With this change, the folder name for the test fixtures (`tests/integration/scrapers/testcache`) are used for tests.

### Big changes

* `docs/sources.md` is a big change
* the files `tests/integration/scrapers/scraper-*-test.js` are pretty
  much complete rewrites -- don't bother looking at the diffs in
  GitHub, just view the whole file.
* moved files to integration tests, rewrote for tape

### Misc small changes

* small tweaks in code to allow for testing, reuse libraries
* added `yarn test:unit` and `yarn test:integration` scripts

## Impacts

* if devs change their scrapers, they may be surprised by integration
  test errors.  This is outlined in the docs/sources.md, but we should
  still send out a message to all devs about this.
* `yarn test` is a bit slower, due to file reads and generation.
  `yarn test:unit` will likely be sufficient for most people.

## Remaining work

* we should add more testing assets for all of the existing scrapers,
  to ensure changes don't introduce regression issues.
* The existing scraper tests (in src/shared/scrapers/**/tests) should
  be removed, as they're no longer used.  That's a separate issue,
  separate PR.
* We can investigate using the ideas in this testing code to run the
  scrapers against real cached data.


tests/integration/scrapers/scrapers-all-test.js
```diff
@@ -1,94 +1,172 @@
...
+
+  let result = null;
+  try {
+    process.env.SCRAPE_DATE = date;
+    result = await runScraper.runScraper(scraperObj);
+  } catch (e) {
+    t.fail(`${scraperName} on ${date}, error scraping: ${e}`);
+  } finally {
+    delete process.env.SCRAPE_DATE;
+    get.get = oldGetGet;
```
camjc, line 75: weird naming to have anything like foo.foo

tests/integration/scrapers/scrapers-all-test.js
```diff
@@ -1,94 +1,172 @@
...
-    const scraperName = scraperNameFromPath(testDir); // e.g. `USA/AK`
-    process.env.OVERRIDE_CACHE_PATH = testDir;
+// Mutex
 
-    // dynamically import the scraper
-    // eslint-disable-next-line
-    const scraperObj = imports(join(testDir, '..', 'index.js'));
+// Each test folder modifies global state (get.get, and
+// process.env.SCRAPE_DATE), so we need to make sure that they're run
+// one at a time.  Global state is bad!
```
camjc, line 106: yup. is there a plan to resolve this eventually?

tests/integration/scrapers/scrapers-all-test.js
jzohrab: Yes, that's the codebase though, I didn't want to refactor it. :-)

tests/integration/scrapers/scrapers-all-test.js
jzohrab: I don't think so, just yet.  It really should be passed to the scraper as an argument.

jzohrab:
I think it's ok to leave the few commented console logs.  If I have a problem with this in the future, I'll use them.  :-)  Thanks for reviewing!

jzohrab:
Please don't merge this yet -- I need to verify that scrapers only work with the existing cache.  Thanks!

jzohrab:
Closing this for now, scraper changes cause tests to fail with bad exceptions, and I need to figure out how to lock down the testing files and caching.  Needs more work!

PR #641, opened 2020-04-03
Fix tn headers

chunder:


src/shared/scrapers/USA/TN/index.js
```diff
@@ -118,13 +118,15 @@ const scraper = {
...
   ],
   _good_headers(data) {
-    if (parse.string(data[0][0]) !== 'Patient county name' && parse.string(data[0][0]) !== 'County') {
+    const countyHeader = parse.string(data[0][0]);
+    const validCountyHeaders = ['Patient county name', 'County', ''];
+    if (validCountyHeaders.indexOf(countyHeader) === -1) {
       return false;
     }
-    if (parse.string(data[1][0]) !== 'Positive') {
+    if (['Positive', 'positives'].indexOf(parse.string(data[1][0])) === -1) {
```
lazd, line 126: Could we just do `parse.string(data[1][0]).toLowerCase().includes('positive')` or would that be too imprecise?

lazd: Let's not!

PR #651, opened 2020-04-04
Calculate timezone for scrapers

hyperknot:
## Summary

Adding timezone calculation before scraping. This is needed for @ryanblock 's cache work.

There was no timezone available before scraping.

This calculates a timezone based on static properties available before scraping. Like country and state strings.

## Changes

One more function in run-scrapers. A lot of change inside country-levels, but that is only visible as a version bump in the linked file here.


hyperknot:
`yarn test` works in local dev, for some reason it's failing in GitHub. Failure unrelated to code change, I just changed a state string.

src/shared/lib/geography/timezone.js
```diff
@@ -0,0 +1,27 @@
+import assert from 'assert';
+import iso2Codes from 'country-levels/iso2.json';
+import usStates from '../../vendor/usa-states.json';
+
+// eslint-disable-next-line import/prefer-default-export
+export const calculateScraperTz = async location => {
```
ryanblock, line 6: Any reason this is async? Looks like sync operations to me, unless you're planning on async lookups for !USA?

ryanblock:
This is great. How do we roll this out to !USA? Anything I can do to help?

ryanblock:
As a quick update: @lazd asked if we could pursue higher resolution timezones; some states and regions exist in multiple timezones. For example, Hundspeth and El Paso counties in TX are on Mountain time, while the rest of TX is on Central time.

That means with 24 crawls per day there could be an hour window where we have off by one errors for that limited number of counties. I think that's sufficient but I'm amenable to sparing some cycles to see if we can think of any other approaches that increase the tz resolution to the county level.

src/shared/lib/geography/timezone.js
hyperknot: > unless you're planning on async lookups for !USA?

Yes, `src/shared/lib/geography/country-levels.js getLocationData` is async and will be used for !USA ones.

hyperknot:
@ryanblock @lazd you are totally right about that one. I can add timezone on a county level, but I don't think anything meaningful is happening in +-1 hour at around midnight. I mean a government website is not going to update numbers at 23:00.

On the other hand, if we _really_ want that, how do we implement that? By the time of scraping we really don't know what counties are being returned. One scraper might even return counties from multiple timezones.

What we can do, is after scraping we can calculate a precise timezone for county level. Then we can at least state the right local time for publishedDate and collectedDate. I believe this is probably the right direction to go, scrape witih state level precision, publish with county level precision.

hyperknot:
> This is great. How do we roll this out to !USA? Anything I can do to help?

Basically we need to assign the ISO codes to all locations and then I can add a few more lines for calculating that. We'll be working on that today.

lazd:
Merging now at @hyperknot's request to avoid complications with future merges.

PR #653, opened 2020-04-04
Enforce kebab case file names

camjc:
- new eslint rule
- manually go fix the existing things breaking the rule.


camjc:
Merge https://github.com/covidatlas/coronadatascraper/pull/652 first.

PR #661, opened 2020-04-04
Fix WV

chunder:


src/shared/scrapers/USA/WV/index.js
```diff
@@ -77,43 +77,85 @@ const scraper = {
...
 
-      const county = geography.addCounty(parse.string(items[0]));
-      let cases = items[1];
+        if (cases.slice(-1) !== ')') {
+          continue;
+        }
 
-      if (cases.slice(-1) !== ')') {
-        continue;
+        cases = parse.number(cases.slice(0, -1));
```
jzohrab, line 106: fyi parse.number returns 0 if the arg is empty string ... you might want to check that, and put "null" if it's an empty string.  I noted this in another issue.

src/shared/scrapers/USA/WV/index.js
```diff
@@ -77,43 +77,85 @@ const scraper = {
...
-      });
-    }
+      return counties;
+    },
+    '2020-04-03': async function() {
+      this.url =
+        'https://wabi-us-gov-virginia-api.analysis.usgovcloudapi.net/public/reports/187b4de8-78ef-40be-9510-7fa9a1ef89f2/modelsAndExploration';
+      const options = {
+        headers: { 'X-PowerBI-ResourceKey': '187b4de8-78ef-40be-9510-7fa9a1ef89f2' }
+      };
```
jzohrab, line 124: I guess there's no real way to check that these don't change.

src/shared/scrapers/USA/WV/index.js
```diff
@@ -77,43 +77,85 @@ const scraper = {
...
+        for (const vc of sec.visualContainers) {
+          if (vc.config.indexOf(heading) !== -1) {
+            const config = JSON.parse(vc.config);
+            let { textRuns } = config.singleVisual.objects.general[0].properties.paragraphs[0];
+            textRuns = textRuns.map(tr => parse.string(tr.value));
+            if (textRuns[0] !== heading) {
+              throw new Error('Unknown json layout');
+            }
+            const pieces = textRuns[1].split(',');
+            for (const row of pieces) {
```
jzohrab, line 142: Do you think it's useful to regex the content of pieces, or of row, for sanity?

src/shared/scrapers/USA/WV/index.js
chunder: This is for historical cached content. I'd just as soon leave it alone unless you find there's an actual issue.

src/shared/scrapers/USA/WV/index.js
chunder: If they change, the rest of this scraper will explode. They've been working for ~16 hours  so I'm hoping it's not a session that will expire.

src/shared/scrapers/USA/WV/index.js
chunder: Using a regex on the row might be more pleasant yes. But it'll be putting lipstick on a pig.

I don't know js regex syntax, so I used the hammer I knew.

src/shared/scrapers/USA/WV/index.js
jzohrab: Sounds good!

PR #580, opened 2020-04-01
Add --onlyUseCache CLI flag 

HerbCaudill:
Adds a CLI flag to prevent new http calls; only the existing cache files are used, to make it easier to make repeatable runs for testing purposes.

PR #664, opened 2020-04-04
Add onlyUseCache command-line flag

jzohrab:
## Summary

(Code salvaged from Herb C's PR #580.)

Adds `--onlyUseCache`, to ensure that we stick with cache!

This change opens up the possibility for testing things entirely offline.  If a cache misses, it returns null, and then scrapers have to deal with that.

## Sample run

```
$ yarn start --location "PA, USA" --onlyUseCache -d 2020-04-04
yarn run v1.22.4
$ node src/shared/cli/index.js --location 'PA, USA' --onlyUseCache -d 2020-04-04
⏳ Fetching scrapers...
✅ Fetched 135 scrapers!
✅ All scrapers are valid!
  🐢  Cache miss for https://www.health.pa.gov/topics/disease/coronavirus/Pages/Cases.aspx at coronadatascraper-cache/2020-4-4/6847ef86804ff0c3c03f31ca00d7eb6d.html
  ❌ Error processing PA, USA:  TypeError: $ is not a function
    at Object.scraper (/Users/jeff/Documents/Projects/coronadatascraper/src/shared/scrapers/USA/PA/index.js:181:28)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async runScrapers ...

```

The `TypeError: $ is not a function` is how this scraper deals with a cache miss, and failed fetches.

## Additional notes

* This introduces more global state, but it's probably OK.  I need to open a separate PR to deal with global state during testing.
* I have only tested this and verified manually, I'm not sure how to test this with JS at the moment given the global state issue.  (If I write a simple test with this, it could break other tests referring to the global state.)  I could potentially write a bash script or something that gets called in its own CI step and process, but wasn't sure how to fit that within the project coding style.

## Further thoughts

Could this be used to diagnose how scrapers handle cache misses and empty returns?  They should all throw better messages, not nonsense like `$ is not a function`.

### DIFF
--- a/src/shared/cli/cli-args.js
+++ b/src/shared/cli/cli-args.js
@@ -54,6 +54,7 @@ const { argv } = yargs
     type: 'boolean'
   })
   .option('onlyUseCache', {
+    alias: 'x',
     description: 'Only use cache (no http calls)',
     type: 'boolean'
   })
--- a/src/shared/cli/cli-args.js
+++ b/src/shared/cli/cli-args.js
@@ -56,4 +56,14 @@ const { argv } = yargs
   .help()
   .alias('help', 'h');
 
+if (argv.date) {
+  process.env.SCRAPE_DATE = argv.date;
+} else {
+  delete process.env.SCRAPE_DATE;
+}
+
+if (argv.quiet) {
+  process.env.LOG_LEVEL = 'off';
+}
+
 export default argv;
--- a/src/shared/cli/cli-args.js
+++ b/src/shared/cli/cli-args.js
@@ -53,6 +53,10 @@ const { argv } = yargs
     description: 'Write to dist folder',
     type: 'boolean'
   })
+  .option('onlyUseCache', {
+    description: 'Only use cache (no http calls)',
+    type: 'boolean'
+  })
   .help()
   .alias('help', 'h');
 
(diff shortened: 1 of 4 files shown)


src/shared/cli/cli-args.js
```diff
@@ -53,7 +53,27 @@ const { argv } = yargs
...
+    description: 'Only use cache (no http calls)',
+    type: 'boolean'
+  })
   .help()
   .alias('help', 'h');
 
+if (argv.date) {
+  process.env.SCRAPE_DATE = argv.date;
+} else {
+  delete process.env.SCRAPE_DATE;
```
camjc, line 66: Where are we setting this otherwise?

lazd:
TBH the movement of the logic into `cli` makes total sense to me, and this seems like a great feature. Thanks, @jzohrab!

src/shared/cli/cli-args.js
```diff
@@ -53,7 +53,27 @@ const { argv } = yargs
     description: 'Write to dist folder',
     type: 'boolean'
   })
+  .option('onlyUseCache', {
+    description: 'Only use cache (no http calls)',
```
lazd, line 57: `x` seems like a fine character, let's use that.

```suggestion
    alias: 'x',
    description: 'Only use cache (no http calls)',
```

jzohrab:
Thank @HerbCaudill, he had the initial idea, I just stole his code and fixed conflicts.

jzohrab:
Merging post reviews.  Thanks!

Issue #647, opened 2020-04-04
city, county, state, country need to be human readable, ISO should only be in cityId, countyId, stateId, countryId


lazd:
## Description

Right now, we mix in ISO-3166 Alpha 3 into the `country` field, but it should be the full, human-readable country name instead.

This was a problem with 2020-4-2 data -- people were looking for USA, but it's United States now.

We need to be consistent.

weslawson:
I'm seeing this in the timeseries.csv file, I initially though there was missing data.

lazd:
Fixed by #683 in 2020-4-5 data.

PR #667, opened 2020-04-05
fix: Make countries human readable (#647)

andys1376:
## Summary

Make countries human readable (#647)

Before this change, countries other than United States were shown in the data reports using their alpha-3 code. Now they use the full, human readable name.

image

## Changes

Changed the transformIds function to transform alpha-3 country codes. The function was previously only relying on countryLevels.transformLocationIds, which does not recognize alpha-3 country codes.

## Additional notes

Unfortunately this does not fix the cross check report - the locations there are still not human readable. This would require more work, likely moving the transformIds function to run before the dedupeLocations function, and modifying the findFeatures and findPopulations functions to work with the transformed ids (perhaps by using the id fields like countryId instead of fields like country).

### DIFF
--- a/src/events/processor/transform-ids/index.js
+++ b/src/events/processor/transform-ids/index.js
@@ -1,6 +1,7 @@
 import * as countryLevels from '../../../shared/lib/geography/country-levels.js';
 import * as geography from '../../../shared/lib/geography/index.js';
 import log from '../../../shared/lib/log.js';
+import countryCodes from '../../../shared/vendor/country-codes.json';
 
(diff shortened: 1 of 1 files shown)


hyperknot:
Thanks for this fix. Right now we are workign very hard in making sure that all countries have ids and nice human readable names!

lazd:
Ok so this is close, but we need to get `countryId` right. I had a branch I was working on this already in, so I may close this out.

lazd:
We got this fixed up tonight, thanks for the PR though!

PR #669, opened 2020-04-05
Getting started doc notes

jzohrab:


jzohrab:
Readme changes only, good enough!

PR #674, opened 2020-04-05
feat: Include population density based on GeoJSON region (closes #645)

### LINKED ISSUE #645: Include population density based on GeoJSON region
## Description

This is super useful information, we should calc it for folks.

## Why do you need this feature or component?

Modeling

andys1376:
## Summary

Include population density based on GeoJSON region (#645)

Add populationDensity field to data report that is the population per square meter of the region.

## Changes

Added @turf/area dependency to calculate the area of the geometry in find-features. Then we divide the population by the area in find-populations to get the density.


src/events/processor/find-features/index.js
```diff
@@ -139,9 +140,11 @@ const generateFeatures = ({ locations, report, options, sourceRatings }) => {
       }
     }
 
-    // Store coordinates on location
+    // Store coordinates and area on location
     if (feature.geometry) {
       location.coordinates = turf.center(feature).geometry.coordinates;
+      // Convert area from square meters to square miles
+      location.area = geojsonArea.geometry(feature.geometry) / 2589988.1103;
```
lazd, line 147: I'd say let's use meters.

package.json
```diff
@@ -34,6 +34,7 @@
     "@js-joda/core": "^2.0.0",
     "@js-joda/timezone": "^2.2.0",
     "@lorenzofox3/for-await": "^0.2.1",
+    "@mapbox/geojson-area": "^0.2.2",
```
lazd, line 37: Please use `@turf/area` instead: https://www.npmjs.com/package/@turf/area

src/shared/lib/schemas/location.json
```diff
@@ -16,6 +16,7 @@
     "hospitalized": { "type": "number", "minimum": 0 },
     "discharged": { "type": "number", "minimum": 0 },
     "population": { "type": "number", "minimum": 0 },
+    "populationDensitySqMi": { "type": "number", "minimum": 0 },
```
lazd, line 19: Let's use `populationDensity` and document it as square meters.

andys1376: Where should it be documented?

lazd: @andys1376 I'll handle that in #505

lazd:
Itching to get this in, if you can handle the switch to `@turf` and square meters tomorrow, let me know. Thanks!

hyperknot:
I agree with using turf and meters!

andys1376:
Updated

src/events/processor/find-features/index.js
```diff
@@ -139,9 +139,10 @@ const generateFeatures = ({ locations, report, options, sourceRatings }) => {
       }
     }
 
-    // Store coordinates on location
+    // Store coordinates and area on location
     if (feature.geometry) {
       location.coordinates = turf.center(feature).geometry.coordinates;
+      location.area = turf.area(feature.geometry);
```
lazd, line 145: Think we ought to include `area` in the final output? Any reason it would be useful (besides, of course, calculating `populationDensity`, which we're already doing).

lazd: We can always add it if someone asks.

lazd:
Nice work, glad to have this in the data!

PR #676, opened 2020-04-05
Remove testcache files, long filenames breaking for some people.

jzohrab:
Remove testcache files, as long filenames are causing problems for some devs.  The integration testing code isn't ready anyway, so these files aren't currently used.


jzohrab:
I'm going to merge this, these files are as yet unused.  The test code was supposed to be merged in right afterwards, but it's good we didn't!

PR #677, opened 2020-04-05
Fixes for KS PDF reading

aed3:


src/shared/scrapers/USA/KS/index.js
```diff
@@ -262,15 +262,13 @@ const scraper = {
       const pdfScrape = await fetch.pdf(this.url);
 
       const data = pdfScrape
+        .filter(item => item && item.y > 6 && item.y < 46)
```
lazd, line 265: Wish we didn't have to hardcode this, but if it works, that's fine.

PR #679, opened 2020-04-05
Idaho tableau

chunder:


src/shared/lib/fetch/get.js
```diff
@@ -65,13 +69,20 @@ export const get = async (url, type, date = datetime.old.scrapeDate() || datetim
...
-        // and throw an error later down that won't bring the whole process down.
-        errorMsg = err.toString();
-      });
+      let response;
+      if (method === 'get') {
+        response = await needle('get', url, { cookies, headers }).catch(err => {
+          errorMsg = err.toString();
+        });
+      } else {
+        response = await needle('post', url, args, { cookies, headers }).catch(err => {
```
lazd, line 82:
```suggestion
        response = await needle(method, url, args, { cookies, headers }).catch(err => {
```

lazd: Kinda seems like we could just pass `method` at all times

chunder: But `needle('get', ...)` doesn't take args. Only `needle('post', ...)` does.

PR #689, opened 2020-04-06
Scrape NZ country-level

camjc:
Just starting to scrape New Zealand, not ready yet. For one the iso2's don't seem to line up with the "DHB"s they're publishing. Theres also a bunch of good data over a few tables, would be nice to scrape them all.


hyperknot:
Can we make thoze zone of out iso2's like in Alaska? If not, is there a Shapefile / GeoJSON source for those? It's not the end of the world if not, but then we cannot visualize it on a map.

camjc:
@hyperknot I decided its best to scrape the country-level properly first, then at least we'll have the pages in our cache, and then can address region-level later on.

hyperknot:
@camjc that's a good idea.

PR #694, opened 2020-04-06
Change the caching strategy of geo-tz library

evansiroky:
This changes the caching strategy of the geo-tz library so that data is stored in-memory rather than using the timed-cache library which will result in no timeouts being applied to the data. Also, a comment has been rewritten to not call out the geo-tz library specifically.

### DIFF
--- a/src/events/processor/find-features/index.js
+++ b/src/events/processor/find-features/index.js
@@ -10,6 +10,10 @@ import espGeoJson from '../vendor/esp.json';
 
 const DEBUG = false;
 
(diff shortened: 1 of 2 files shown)


lazd:
Hey, thanks! We were scratching our heads when CI failed after introducing `geo-tz` to the build, but otherwise very happy with the library. Cheers!

evansiroky:
Hmm, I can see how that must have been a little annoying, especially when you're using it to process data and you'd expect your script to end immediately. I'll probably change the default behavior of the library as a result.

Issue #687, opened 2020-04-06
Missing population for Arunachal Pradesh, Assam, Jharkhand India


lazd:
## Location, date, and short issue description

  ❌ Arunachal Pradesh, iso1:IN: ?
  ❌ Assam, iso1:IN: ?
  ❌ Jharkhand, iso1:IN

pankaj-giri:
@lazd 
Can you tell me how this file gets populated - ./src/events/processor/vendor/world-states-provinces.json and what is the source for the information?

hyperknot:
@pankaj-giri we'd need to convert it to iso codes, have a look at Italy for example. Then all of these would be automatically populated.

PR #702, opened 2020-04-07
Updated the population data for some of the missing states in India

pankaj-giri:
## Summary

#687

Some of the states were missing in the json.

Added the missing states in the json

### DIFF
--- a/src/shared/scrapers/ID/index.js
+++ b/src/shared/scrapers/ID/index.js
@@ -1,0 +1,38 @@
+import assert from 'assert';
+import * as fetch from '../../lib/fetch/index.js';
+import * as parse from '../../lib/parse.js';
+import maintainers from '../../lib/maintainers.js';
+import getKey from '../../utils/get-key.js';
+
+const labelFragmentsByKey = [{ recovered: 'sembuh' }, { deaths: 'meninggal' }, { cases: 'positif covid-19' }];
+
+const scraper = {
+  country: 'iso1:ID',
+  maintainer: [maintainers.camjc],
+  priority: 1,
+  sources: [
+    {
+      description: 'Ministry of Health Republic of Indonesia',
+      name: 'Ministry of Health Republic of Indonesia',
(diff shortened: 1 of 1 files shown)


hyperknot:
Thanks for doing this, however we are migrating away from using this json file. Can you check if the population information is correct on this page?
https://github.com/hyperknot/country-levels/blob/master/docs/iso2_list/IN.md

pankaj-giri:
It looks very different from - 
https://en.wikipedia.org/wiki/List_of_states_and_union_territories_of_India_by_population
The JSON file earlier was referring to this link too.

hyperknot:
Can you point to one row which is very different? If it's only about not-the-most-recent census count, then we should update the Wikidata values.

pankaj-giri:
If you take a look at the first entry 
Andaman and Nicobar Islands - wikipedia link lists the population as 380,520 whereas IN.md lists it as 356,152.
Andra Pradesh - wikipedia - 49,386,799, IN.md - 84,580,777. Thats quite a variance.
Looks like what you're picking from wikidata is very dated - 2005

pankaj-giri:
https://en.wikipedia.org/wiki/List_of_states_and_union_territories_of_India_by_population
You can use this one which is from the 2011 census

pankaj-giri:
Non related to this problem - we have the scrapers run at the individual level broken I think.. Eg.
yarn start --location "IND" or "IN" does not work. This was working yesterday. Same is the case with 
yarn start --location "ITA" or "IT"

hyperknot:
@pankaj-giri that Andaman and Nicobar Islands  is just old data. 
Andra Pradesh on the other hand is very complicated and I'm chatting with Wikidata people about it.

hyperknot:
@pankaj-giri I'm working on fixing the state scrapers.

lazd:
@pankaj-giri run `yarn start -l "iso1:IN"`

PR #709, opened 2020-04-08
[patch] support deaths in table now, backwards compatible

camjc:
They have 'confirmed cases' in their deaths row as a parenthetical, so the unknown row assertion didn't trigger, haha.
image

Keep the paragraph scraper as a fallback for backwards compatibility.


camjc:
@lazd approves this message … or … merge.

PR #710, opened 2020-04-08
India using ISO codes

hyperknot:
## Summary

India migrated to using ISO codes. Population is present on all of them.

@pankaj-giri can you check it out?

Corona Data Scraper 2020-04-08 04-36-41


pankaj-giri:
This is awesome @hyperknot. Thanks for letting me know.

Issue #631, opened 2020-04-03
Maine data way off


kb1ujs:
Data for Maine is way off on number of deaths.  Looks like maybe hospitalizations are being counted instead as deaths.

kb1ujs:
I looked at the data closer and that's exactly what's happening.  The death data is not in the file and the column marked deaths is actually hospitalizations.

kb1ujs:
Table format on Maine CDC website changed on April 2, breaking the scraper.  Will try to submit a pull request to fix.

lazd:
Thanks @kb1ujs, please do send along a PR!

jzohrab:
Hi @kb1ujs , thanks for the PR.  This was recently merged, so if this problem is resolved in the next release or so please close this issue.  Thanks again! jz

praging:
three dates have not yet been fixed: 03-30, 03-31, and 04-01

kb1ujs:
Looks like the format change occurred a few days earlier than I thought.  I'll fix it and submit a new pull request.

lazd:
Closed by #718

PR #718, opened 2020-04-08
Adjust start date for Maine CDC Website format.

kb1ujs:
Fixes March 30 - April 1 data as mentioned in covidatlas/coronadatascraper#631,


lazd:
Thanks for healing the timeseries!

Issue #706, opened 2020-04-07
Yarn Flags (Documentation Fix and Feature Request)


praging:
yarn flags have been changed over the past week or so -- need to be explained more fully in the [documentation](https://github.com/covidatlas/coronadatascraper/blob/master/docs/getting_started.md).

add: the `-c` flag for country-specific scraper runs
remove: the `-l` flag for location-specific scraper runs (the Ventura County example no longer works)

the latter should be implemented in a similar way and documented -- current way to run a region-specific scrape is "quite ugly" (@hyperknot)

the `-l` flag should also probably be renamed now that it is not used for location

hyperknot:
I mean `yarn start -l "Ventura County, iso2:US-CA, iso1:USA"` is not exactly nice and also it doesn't even work now.

jzohrab:
Ventura county has the following:

```
const scraper = {
  county: 'Ventura County',
  state: 'CA',
  country: 'iso1:US',
```

Should you be using the iso2 there, or should it be `Ventura County, CA, iso1:US` ?

hyperknot:
`yarn start -l "Ventura County, CA, iso1:USA"` doesn't work either. I think the confusing part is that -l is also trying to match on filename, and that combination of filename matching and location names is not really working.

jzohrab:
I think we should match on filename only, it's easier to grok when quickly running one scraper, or a group.  It could be a regex, and the match would be checking against the path relative to the scrapers root, all parts joined by '/'.  So, '.*ventura.*' would find anything with 'ventura' in the path, '^US' anything starting with US, '.*index.*' would run pretty much everything.
Thoughts?

hyperknot:
I totally agree we should match on filename/subdir. So we should be able to do US/CA or `US/CA/ventura` or something similar.

lazd:
Yeah these flags are unpredictable and confusing, I know because I made them that way as a means to an end :)

Let's make them not suck! Matching on filename/subdir is GREAT.

jzohrab:
Assigning to self: I'm going to do this now, it should be a small fix and will help everyone, including me.

jzohrab:
PR opened. #720

lazd:
`yarn options` now clues you into all the deets. Cheers!

PR #720, opened 2020-04-08
Fix command line scraper filters

jzohrab:
## Summary

Updates the `--skip` and `--location` filters to instead use relative file paths only.  We're moving to iso codes, so the filters are changing, and it's hard to guess what the right thing is.

Addresses issue #706 (yarn flags), and docs.

## Sample calls

```

$ yarn start --location "US/NV"
yarn run v1.22.4
$ node src/shared/cli/index.js --location US/NV
⏳ Fetching scrapers...
✅ Fetched 138 scrapers, filtered to 3.
... etc


$ yarn start --location "US/NV/carson-city.js"
yarn run v1.22.4
$ node src/shared/cli/index.js --location US/NV/carson-city.js
⏳ Fetching scrapers...
✅ Fetched 138 scrapers, filtered to 1.
... etc
```

`start` quits if nothing matches:

```
$ yarn start --location "xxx"
yarn run v1.22.4
$ node src/shared/cli/index.js --location xxx
⏳ Fetching scrapers...
location filter returned 0 scrapers.  Please check docs/getting_started.
No sources, quitting.
```

I didn't bother using regexes, because it seemed that there were too many chances to run the wrong thing ... e.g., "US" matches "RUS". A scraper named 'carson-city.js' exists, but this won't run it:

```
$ yarn start --location "carson"
yarn run v1.22.4
$ node src/shared/cli/index.js --location carson
⏳ Fetching scrapers...
location filter returned 0 scrapers.  Please check docs/getting_started.
No sources, quitting.
```

## Changes

* Moved scraper filtering code up to the loading of the filters.
* Quick exit during generation if there are no sources.
* Updated docs with examples.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -70,9 +70,9 @@ To run the scrapers for today:
 yarn start
 ```
 
-### Run scrapers for only one location
+### Run selected scrapers
 
-To scrape just one location, use `--location`/`-l`
+To use a subset of scrapers, use `--location`/`-l`
 
 ```
 yarn start --location "US/PA"
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -82,9 +82,11 @@ The `location` value should match a path under `src/shared/scrapers/`.
 
 Examples:
 
-* `US`: run all U.S. scrapers
-* `US/CA`: run all California scrapers
-* `US/CA/alameda-county.js`: run this single scraper
+* `yarn start --location "US"`: run all scrapers in `src/shared/scrapers/US` and child folders
+* `yarn start --location "US/CA"`: run all scrapers in `src/shared/scrapers/US/CA` and child folders
+* `yarn start --location "US/CA/alameda-county.js"`: run this single scraper
+* `yarn start --location "AU"`: run all scrapers in `src/shared/scrapers/AU` and child folders
+* `yarn start --location "AU/index.js"`: run this single scraper
 
 
 ### Skipping a scraper
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -43,7 +43,7 @@ If you get an error message saying you have an incompatible version of
 node versions: [install](https://www.npmjs.com/package/n) it and run
 `n lts`.
 
-### 3. Run the scraper
+### 3. Run the scrapers
 
 ```
 yarn start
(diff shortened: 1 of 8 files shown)


hyperknot:
I think -c country would be important, otherwise there is no way to test Australia. The -i / id on the other hand should be removed, I believe (I added it last week). Basically instead of -i we can specify the JS file.

jzohrab:
@hyperknot, couldn't you test Australia with `yarn start --location "AU"` ?

jzohrab:
Clarified with @hyperknot over Slack.  `yarn start --location "AU"` runs all scrapers in Australia.  `yarn start --location "AU/index.js"` runs just that single file.  I _think_ this is clear in the docs. :-)

hyperknot:
I think AU is a good example for state overrides and definitely not trivial, I'd include that one in the docs, including specifying AU/index.js.

jzohrab:
Updated docs, thx @hyperknot .

jzohrab:
Note to self/merger: when this is merged, we should send out a message: "Hi all, we've changed the --location command line switch to use folder paths instead.  e.g., to run all scrapers in src/shared/scrapers/US/NV, you'd specify "--location US/NV" .  See docs/getting_started.md for more examples.

hyperknot:
Looks good to me! Agree about the note.

src/events/crawler/get-sources/load-sources.js
```diff
@@ -1,6 +1,25 @@
...
+function includeLocation(location, opts) {
+  const relpath = location._path.replace(/.*?src.shared.scrapers./, '');
+
+  if (opts.skip && relpath.startsWith(opts.skip)) return false;
+
+  if (opts.location && !relpath.startsWith(opts.location)) return false;
+
+  // select location based on country level id
+  // for example "yarn start -i id3:AU-VIC"
+  if (opts.id && opts.id !== countryLevels.getIdFromLocation(location)) return false;
```
hyperknot, line 19: So I'd remove this id. I added it and there is clearly no point of doing this one anymore.

hyperknot:
Also, please remove from src/shared/timeseries/run-crawler.js as well.

src/events/crawler/get-sources/load-sources.js
jzohrab: Got it, removing, and checking timeseries run-crawler as well.

jzohrab:
Updated, thanks @hyperknot .

docs/getting_started.md
```diff
@@ -70,26 +70,31 @@ To run the scrapers for today:
 yarn start
 ```
 
-### Run only one scraper
+### Run scrapers for only one location
```
hyperknot, line 73: So this is not "only one location" anymore, but this is the last minor thing I could find.

hyperknot:
Basically perfect, I added one more note.

docs/getting_started.md
jzohrab: Yeah I dithered on that one ... is the US one location?  Actually, I'll call it "Run selected scrapers."  Thanks!

jzohrab: Updated, thank you!

jzohrab:
All @hyperknot comments taken into account.  Awaiting further feedback, but if I don't hear any I'll merge this tomorrow and notify the team.

Issue #633, opened 2020-04-03
Scraper for Argentina


shaperilio:
https://www.argentina.gob.ar/coronavirus/informe-diario

This is from the federal government. They are publishing two PDFs per day. "Vespertino" = evening, "Matutino" = morning. They're probably meeting minutes.

Pros:
 - They maintain previous days' files online (but we should start caching just in case).
 - Later PDFs have cases tabulated by province.

Cons:
 - PDFs
 - Inconsistent filenames (must parse HTML links to get PDFs)
 - Additional information / data in paragraph form

shaperilio:
I've made an initial attempt at this in my [argentina](https://github.com/shaperilio/coronadatascraper/tree/argentina) branch. Need help, because what i'm doing breaks caching.

PR #721, opened 2020-04-08
Ukraine

ciscorucinski:
## Summary
Adding Ukraine scraper

## Changes
- Added scraper
- Added me as a maintainer

## Additional notes
None


src/shared/scrapers/UKR/index.js
```diff
@@ -0,0 +1,99 @@
...
+  'Kharkiv region': 'iso2:UA-63',
+  'Kherson region': 'iso2:UA-65',
+
+  'Khmelnytskyi region': 'iso2:UA-68',
+  'Khmelnytsky region': 'iso2:UA-68',
+
+  'Kirovohrad region': 'iso2:UA-35',
+  'Kirovograd region': 'iso2:UA-35',
+
+  Kyiv: 'iso2:UA-30',
```
hyperknot, line 24: Kyiv would never be used, would it?

src/shared/scrapers/UKR/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+  'Lviv region': 'iso2:UA-46',
+  'Mykolaiv region': 'iso2:UA-48',
+
+  'Odesa region': 'iso2:UA-51',
+  'Odessa region': 'iso2:UA-51',
+
+  'Poltava region': 'iso2:UA-53',
+  'Rivne region': 'iso2:UA-56',
+  Sevastopol: 'iso2:UA-40',
```
hyperknot, line 41: Also, isn't it "Sevastopol region"?

hyperknot:
Thanks for this, it looks amazing! 

1. Can you rename the folder to UA?

hyperknot:
image

src/shared/scrapers/UKR/index.js
ciscorucinski: For all the duplicate values, the top one comes from the list you gave me before, the second one comes from what is in the JSON

hyperknot: The list I gave is not important, only the ids on it. We just need to map to the ids. You can even do `.replace(/region/g, '')`and just make everything shorter.

src/shared/scrapers/UKR/index.js
ciscorucinski: looking into this. I think some regions aren't included. I thought there was a mismatch between the list you gave me and what the JSON had.

ciscorucinski: `iso2:UA:40` is a city https://en.wikipedia.org/wiki/ISO_3166-2:UA

hyperknot:
So actually I'd recommend:
1. Removing all the following logic:
```
region.label.en === 'Kyiv'
            ? geography.addCounty(region.label.en, 'city')
            : geography.addCounty(region.label.en, 'region');
```
2. Using the exact `region.label.en` strings in the mapping.
3. Adding a console.error() + continue when a region appears which is not in the map.

ciscorucinski:
I don't need to have the `geography.addCounty(region.label.en, 'region');` code at least in one place?

hyperknot:
I'll test it out if I can get it done without the geography.addCounty bits.

hyperknot:
@ciscorucinski I pushed a commit to the PR. It explains what I meant by simplifying the code. Names really don't matter here, only the ids have to be correctly identified. If you are ok with the changes I can merge this tomorrow.

hyperknot:
image

lazd:
This looks great! Is the Ukraine source a timeseries? If so, there is a special way to handle it.

lazd:
Gonna merge this as-is because it's awesome, but let's try to figure out if their api has a `from` field where we can get a full timeseries.

lazd:
Gonna merge this as-is because it's awesome, but let's try to figure out if their api has a `from` field where we can get a full timeseries.

I'll make a few changes to the way `this.url` is used and fix caching (should not pass `false` as second argument), but this is gold. Thank you!

lazd:
I now see why you pass `false` -- you can't go back in time for the timeseries otherwise, of course. But the URL bit needed to not modify `this.url` with the date, or we append the date for each date in the timeseries as we run (not true since we moved to separate processes, but we want to work under the assumption that the module is not re-imported for each day).

shaperilio:
> I now see why you pass `false` -- you can't go back in time for the timeseries otherwise, of course. But the URL bit needed to not modify `this.url` with the date, or we append the date for each date in the timeseries as we run (not true since we moved to separate processes, but we want to work under the assumption that the module is not re-imported for each day).

You're referring to the `false` in the `const data = await fetch.json(this.url, false, { disableSSL: true })` call? Agreed, we don't want this because it means it never uses the cache - this is what I meant by having a need to "back-fill" cache. This source is a good example of that; [Argentina](#633) also has previous data up and would also need this type of functionality.

shaperilio:
I just tested this, it *does* use the cache, and it can "go back in time", so disregard what I wrote.

Issue #703, opened 2020-04-07
Add timeseries scraper for Washington, D.C.


lazd:
## Location name

Washington, D.C., USA

## Source URL

https://microstrategy.dc.gov/MicroStrategyLibrary/app/E52A533B11EA01A28FE80080EFC53DA4/30FE682111EA6FB5A6620080EF357B1F/K53--K46??ui.navigation.share=false&ui.navigation.gotoLibrary=false&ui.navigation.options=false&ui.navigation.filter=false&ui.navigation.reset=false&ui.navigation.bookmark=false&app.embedded=true&auth.freeLogin=false&ui.navigation=true&ui.navigation.title=true&ui.navigation.toc=true&ui.message=false

## Notes/comments

Requires a token

aed3:
I can look into it

jzohrab:
Fixed in #724, afaict.  Cheers!

PR #724, opened 2020-04-09
Fixed DC to use time series CSV data

aed3:
First, goes to a website to retrieve the cookie necessary to be authorized for the CSV exporter form.
Then, grabs a value from that form which is needed in the exported CSV's USL. It's different for every time there's a different cookie.
Next, retrieves the CSV and parses it after removing a title line and extra characters.
Last, finds the correct column for the date we're looking for.
#703 
Well, at least I learned a lot from this mess...


lazd:
holy shit wow nice work!

PR #727, opened 2020-04-09
Fixed MA scraper

aed3:
They split Dukes and Nantucket and removed the "EditDate" field


lazd:
Is this backwards compatible? Seems like a big change that needs to be behind a date?

aed3:
Yes, entirely. Triple checked it this time.

lazd:
THANK YOU!!!

Issue #637, opened 2020-04-03
datetime.now.utc() should have a Z at the end


lazd:
**Description**
/title

**To Reproduce**
1. Run `datetime.now.utc()`
2. Observe Z-lessness

**Expected behavior**

`datetime.now.utc()` be has the Z

**Screenshots**
![image](https://user-images.githubusercontent.com/201344/78409812-8348a180-75bf-11ea-9830-c7f261901a7c.png)

**Additional context**
It's true

hyperknot:
Right, I've just checked Wikipedia and it says:
```
If no UTC relation information is given with a time representation, the time is assumed to be in local time. 
```

So we either need to add Z or `+00:00` to specify UTC.

jzohrab:
Added a fix to close one issue.

PR #731, opened 2020-04-09
Fix issue 637: add Z to datetime.now.utc()

jzohrab:
## Summary

Fix for #637 

Before the change, datetime.now.utc() didn't have Z.
After the change, it does.

## Additional notes

@hyperknot - is this all there is to it?  I ran `yarn start` and didn't see any obvious explosions.


hyperknot:
Looks good to me!

PR #734, opened 2020-04-10
remove extraneous args and imports

camjc:
This stuff was going nowhere, so lets delete it to stop being misleading.


src/events/crawler/scrape-data/normalize-locations.js
```diff
@@ -1,13 +1,11 @@
 import path from 'path';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
camjc, line 2: Much better to scope these. Linter would have caught that the country-codes wasn't used if we'd done it in the first place.

qgolsteyn:
LGTM :+1: thanks for this cleanup!

PR #741, opened 2020-04-10
Fix France

qgolsteyn:
## Summary

This PR updates the French scraper, moves to use the data source from data.gouv.fr (an official datasource, away from the OpenCOVID-fr, which is a community project that relies on manually scraping the official datasource).

It adds a county granularity to the data (departements). It uses ISO2 to identify regions and departments. 

A couple of important consideration:
- they don’t have a cases number for regions, only a national one. It is unclear if this number is for metropolitan france or the whole country. They share testing info for each region, but the positive tests only account for 20% of the country total
- they only provide hospital deaths regionally, the national death number includes more than that, but that data is not available
- they don’t have number of recovered, only number of discharged patients
- they only have number of _current_ patients in hospital, so I had to do some math to get the number of cumulative patients 
- overseas territories have their own country code, meaning JHU treats them as a different country. To avoid double counting the same place, I aligned the French dataset to use country codes for these territories.
- data is delayed by a couple of days (~3 days). The way the scraper now works is if we are requesting data for a particular day (say 04/03), we will create a cumulative sum up to that day. This has the effect of making data appear to stagnate in the last 3 days before today (as no info is currently available).

**Currently missing an ISO code for a departement (iso2:FR-10) and the population of Paris (iso2:FR-75).** @hyperknot is looking into it.


src/shared/lib/transform.js
```diff
@@ -91,7 +91,7 @@ export const transposeTimeseries = function(timeseriesByLocation) {
   return { timeseriesByDate, locations };
 };
 
-const caseFields = ['cases', 'recovered', 'active', 'deaths', 'tested'];
+const caseFields = ['cases', 'recovered', 'active', 'deaths', 'tested', 'hospitalized', 'discharged'];
```
hyperknot, line 94: Isn't discharged the same as recovered?

qgolsteyn: No, discharged refers to discharged from the hospital. Patients may still be sick and recovering at home. The number of recovered is not available for France.

src/shared/lib/geography/country-levels.js
```diff
@@ -51,43 +51,50 @@ export const getLocationData = async id => {
...
   const locationData = await getLocationData(id);
-  if (Array.isArray(locationData)) {
-    const features = await Promise.all(locationData.map(l => getFeature(l.countrylevel_id)));
-    return geography.combineFeatures(features);
-  }
-  if (locationData.geojson_path) {
-    const geojsonPath = path.join(countryLevelsDir, 'geojson', locationData.geojson_path);
-    const feature = await readJSON(geojsonPath);
-    return feature;
+  if (locationData) {
```
hyperknot, line 63: I'd invert this for cleaner code. If not: return undefined, ...
Also, should we settle on null or undefined?

src/shared/lib/geography/country-levels.js
```diff
@@ -51,43 +51,50 @@ export const getLocationData = async id => {
...
 };
 
 export const getPopulation = async id => {
   const locationData = await getLocationData(id);
-  if (Array.isArray(locationData)) {
-    return locationData.reduce((a, l) => {
-      a += l.population;
-      return a;
-    }, 0);
+  if (locationData) {
```
hyperknot, line 79: What if one if the array is [valid, valid, undefined, valid]?

hyperknot:
Thanks for this, it looks like really nice work! I'm thinking about the whole country-levels.js change, and so far I think maybe it's better with that assert. Introducing the if-s introduce a lot of edge cases.

Why not just test for the existance of the iso2 code by doing 
```
import iso2Codes from 'country-levels/iso2.json';
```
in the scraper?

qgolsteyn:
Can do! I removed that assert to complete my work, but I can revert back once the data is up to date in `country-level`. I'm okay with allowing the crawler to fail here.

hyperknot:
I think scrapers should either use manual mapping, like IT or UA wherer there is no risk, or dynamic mapping with a `code in iso2codes` lookup to make sure it exists. That way we can handle everything much nicer in the scraper level, sometimes it's a return, sometimes it's a continue or a filter.

For now, I'd recommend to skip the -10 region, until I can fix the dataset. In Italy I used '-' to mark skipped regions and then later I filter to `!== '-'`.

qgolsteyn:
I kept the FR-10 department to ensure aggregated data are still accurate. I simply removed the `iso` prefix.

hyperknot:
Can you add an import iso2codes check?

hyperknot:
```
import iso2Codes from 'country-levels/iso2.json';
```

qgolsteyn:
Added a check. Let's wait to fix `country-level` before merging this PR.

hyperknot:
Yes, that's nice logice! What do you think about skipping instead of doing an assert?

qgolsteyn:
Skip will result in wrong aggregated values. I would prefer failing rather than pushing incorrect data.

hyperknot:
That's right! But then everything might be broken in the future unless someone fixes it.

hyperknot:
I fixed the missing FR-10, it'll be in the next country-levels release, which I'll do tomorrow morning.

lazd:
@hyperknot we should wait to merge this until tomorrow?

hyperknot:
@lazd yes, there are asserts now.

Updates to country-levels:
- added BE-WHT, FR-10
- fixed FR-75, Paris

It'll be uploaded soon.

lazd:
@hyperknot merge at will once it's ready!

hyperknot:
This is working now with the new country-levels, however I'm puzzeled by the data here. Do we really have abolutely no case information? 

Also we have to do something with this?
```
['cases', 'recovered', 'active', 'deaths', 'tested', 'hospitalized', 'discharged'];
```

I don't think we can should the architecture to soo many fields, or do we want to? CSV all need to be updated + I think JSONs should be ok + documentation. @lazd @ryanblock

All in all, I'm really puzzled that we don't have case data in France, one of the Top 5 countries with COVID. Isn't there a different data source? I mean what can we do like this?

The map will always look like a big hole for France.

image

hyperknot:
I think at least for the country level we should have the case number, it seems to be available.

qgolsteyn:
We've been supporting hospitalized and discharged, even if we currently don't output it in our CSV files. See [data_fields](https://github.com/covidatlas/coronadatascraper/blob/master/docs/data_fields.md).

There are no cases number regionally. Have been looking for them for a while now. [Wikipedia](https://fr.wikipedia.org/wiki/Pand%C3%A9mie_de_Covid-19_en_France) and the [French Covid Project](https://github.com/opencovid19-fr/data/issues/435) have both mentioned that this data is no longer provided by the French public health authority.

Another scraper could be provided for national numbers. Or we could use the number from JHU.

hyperknot:
OK, we can definitely use JHU for national numbers. I think this is good to be merged, if 'hospitalized', 'discharged' is OK by the team.

PR #744, opened 2020-04-11
fix: data accuracy (re-align table columns)

ArnavD74:
## Summary

The data source re-ordered the number of columns they have in their data table, so I fixed the right columns to use. No issue relation.

Cases/deaths/recovered figures were not accurate.

Cases/deaths/recovered figures are now even with data source.

## Changes

Just changed which columns are being used for what.

### DIFF
--- a/src/shared/scrapers/IN/index.js
+++ b/src/shared/scrapers/IN/index.js
@@ -41,7 +41,7 @@ const countryLevelMap = {
 
 const scraper = {
   country: 'iso1:IN',
-  url: 'https://www.mohfw.gov.in/', // dashaputra.com/goi
+  url: 'https://www.mohfw.gov.in/',
   type: 'table',
   aggregate: 'state',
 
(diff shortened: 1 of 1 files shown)


jzohrab:
Hi @ArnavD74 , the change looks good, but do you know if the change works with any files that might be cached?  If the cache contains data with the columns switched, then this change will erroneously change past data as well!

src/shared/scrapers/IN/index.js
```diff
@@ -77,8 +77,8 @@ const scraper = {
       const data = {
         state: stateMapped,
         cases: parse.number($tr.find('td:nth-child(3)').text()),
-        deaths: parse.number($tr.find('td:nth-child(6)').text()),
-        recovered: parse.number($tr.find('td:nth-child(5)').text())
+        recovered: parse.number($tr.find('td:nth-child(4)').text()),
+        deaths: parse.number($tr.find('td:nth-child(5)').text())
```
jzohrab, line 81: I believe that these column re-orderings may _not_ be backwards compat with cached data ... if the columns were in a different order in the past.  See the docs at https://github.com/covidatlas/coronadatascraper/blob/master/docs/sources.md#what-to-do-if-a-scraper-breaks .  If I'm wrong, LMK!  Thank you very much for the contribution! jz

jzohrab:
Thank you @ArnavD74 , we have an alternative which will be backwards compatible, in #745.  Regards, and I hope you come back to contribute again! jz

PR #745, opened 2020-04-11
fix iso1:IN India state and country level scraper

camjc:
## Summary
India moved some columns around. In master we look for columns by their order, now we'll look by their heading content.
In place of https://github.com/covidatlas/coronadatascraper/pull/744/files which wouldn't be backwards compatible.

### DIFF
--- a/src/shared/scrapers/IN/index.js
+++ b/src/shared/scrapers/IN/index.js
@@ -1,9 +1,16 @@
 import assert from 'assert';
 import * as fetch from '../../lib/fetch/index.js';
 import * as parse from '../../lib/parse.js';
+import * as transform from '../../lib/transform.js';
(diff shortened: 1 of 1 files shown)


jzohrab:
LGTM.  @camjc does this work correctly with the data in the cache, and present day?  Cheers!

camjc:
As far as I can tell, please QA me though.

Idea is it reads the headings, figures out what things they mean, and then uses that order to build up the data.
That way they can reorder all they like and we won’t care.

jzohrab:
Will get to it tomorrow, if someone else doesn't.  Cheers and thanks @camjc .

lazd:
Sounds good to me, if it doesn't work, we'll fix tomorrow! Merging ahead of tonight's release.

PR #746, opened 2020-04-11
feat: Belgium scraper

qgolsteyn:
This PR adds a scraper for Belgium, scraping from Belgium's official source. It provides the following info:
- cases provincially
- hospitalized/discharged provincially
- deaths regionally
- tested nationally

**Missing ISO code BE-WHT.** @hyperknot was notified. Can be merged once this code is added to `country-levels`


hyperknot:
Updates to country-levels:
- added BE-WHT, FR-10
- fixed FR-75, Paris

I'll be uploaded soon.

lazd:
@hyperknot merge at will once it's ready!

src/shared/scrapers/BE/index.js
```diff
@@ -0,0 +1,136 @@
...
+      }
+
+      nationalData = transform.sumData([regionData], nationalData);
+      data.push(regionData);
+    }
+
+    data.push(nationalData);
+
+    for (const item of data) {
+      assert(iso1Codes[item.country.replace('iso1:', '')], `Missing iso1 code for country ${item.country}`);
```
hyperknot, line 124: Error processing iso1:BE:  TypeError: Cannot read property 'replace' of undefined
    at Object.scraper (/Users/user/nonsync/dev/newcases-org/covidatlas/src/shared/scrapers/BE/index.js:124:37)

hyperknot:
This works great!

One issue is that Brussels should be added both as a county and a state, to make sure that the maps are covering both layers properly / adding up counts correctly.

image

The other is the tiny issue about the country assert line, I recommend removing it. Since country is fixed there is no point checking it.

PR #754, opened 2020-04-11
feat: South Africa scraper

qgolsteyn:
This PR adds a scraper for South Africa, based on the [covid19za](https://github.com/dsfsi/covid19za) dataset. We scrape number of cases and deaths regionally, and number of tests nationally. The data collected was crossed checked with Worldometer and Wikipedia for accuracy.


hyperknot:
Great!

PR #758, opened 2020-04-12
fix iso1:KR South Korea state and country level scraper

camjc:
## Summary
Change to not care about order.
Use the iso codes for states.


hyperknot:
Nice!

PR #760, opened 2020-04-12
fix iso1:VI United States Virgin Islands

camjc:
## Summary
Make Virgin Islands more robust, use some utils to have less code, yay.


lazd:
Looks like it fails for `2020-4-4` and `2020-4-5` :'(

```
yarn timeseries -d 2020-4-4 -l VI
```

PR #761, opened 2020-04-12
fix iso1:PR Puerto Rico

camjc:
## Summary
Seems like main issue was lack of `www.` for some reason isn't resolving. While I'm here I figured I'd get rid of the order-dependence.


lazd:
Nice! 

No we just need to stuff the number parser in there too so we can parse `twenty` and stuff...

camjc:
@lazd huh? Sorry, I don’t follow your comment?

PR #762, opened 2020-04-12
fix US CA Los Angeles

camjc:
- Interestingly LA is replacing client-side using
  a small JS file that looks for IDs.
- Grab the JS file, find the JSON, then use
  that directly.


lazd:
Yay this works thanks!

PR #764, opened 2020-04-12
fixed iso1:UA incorrect usage of county

ciscorucinski:
## Summary
Modified Ukraine to use `state`  rather than `county` level. This was a copy-paste mistake that wasn't caught.

## Changes
Changed `county: clIdMap` to `state: clIdMap`

### DIFF
--- a/src/shared/scrapers/UA/index.js
+++ b/src/shared/scrapers/UA/index.js
@@ -70,7 +70,7 @@ const scraper = {
       }
 
       regions.push({
-        county: clId,
+        state: clId,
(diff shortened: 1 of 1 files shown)


lazd:
@camjc it's actually a timeseries, and we don't have cache for prior days, so not passing the date is required -- it will hit the server once for every day in the timeseries

But yea, I think it would work with `certValidation: false` on the scraper object.

camjc:
Do you have an example of when and when not to have that second (date) arg set? I’m having a hard time understanding it’s use. I think I *almost* understand it haha.

PR #772, opened 2020-04-12
Convert Polygons to MultiPolygons for concatenating counties

shaperilio:
Reproduce the problem with `yarn start -l US/UT`. It will crash.

Some counties in usa-counties.js are of type `Polygon`; simply wrapping their coordinates in an array makes them into `MultiPolygons`, and then this concatenation will work.

### DIFF
--- a/src/events/processor/find-features/index.js
+++ b/src/events/processor/find-features/index.js
@@ -217,7 +217,7 @@ const generateFeatures = ({ locations, report, options, sourceRatings }) => {
       const clId = countryLevels.getIdFromLocation(location);
       if (clId) {
         const feature = await countryLevels.getFeature(clId);
-        storeFeature(feature, location);
+        if (feature !== null) storeFeature(feature, location);
         continue;
       }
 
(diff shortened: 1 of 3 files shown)


src/shared/lib/geography/country-levels.js
```diff
@@ -74,6 +75,10 @@ export const getFeature = async id => {
       },
       geometry: newGeometry
     };
+    if (newGeometry.coordinates.length === 0) {
+      console.error(`Combined geometry for ${newFeature.properties.name} (${id}) does not contain any coordinates!`);
```
hyperknot, line 79: Can you make this an assert?
We should not pass on an empty geometry, it's a bug in country-levels if one is empty.

src/events/processor/find-features/index.js
```diff
@@ -217,7 +217,7 @@ const generateFeatures = ({ locations, report, options, sourceRatings }) => {
       const clId = countryLevels.getIdFromLocation(location);
       if (clId) {
         const feature = await countryLevels.getFeature(clId);
-        storeFeature(feature, location);
+        if (feature !== null) storeFeature(feature, location);
```
hyperknot, line 220: We don't have to care for this, it should throw an exception.

src/shared/lib/geography/country-levels.js
shaperilio: Addressed.

hyperknot:
Thanks for this @shaperilio. I added one more assert and cleaned up the code a bit, now it's good.

PR #773, opened 2020-04-12
European countries round 1

qgolsteyn:
This PR adds the following countries:

- Austria
- Czechia
- Cyprus
- Estonia
- Ireland
- Netherland
- Poland
- Norway

It also adds some TODO files containing details to scrape Latvia, Estonia, Romania, Sweden, Scotland and Portugal, which will be included in a future PR but kept in this PR for reference.

All countries have associated tests, and include timeseries data.


hyperknot:
This is amazing, 8 new scrapers, 35 new files, all using ISO codes!

hyperknot:
Did you check each country on the map with `yarn dev`?

qgolsteyn:
Yup, ran timeseries for each country + did one run with the entire dataset

qgolsteyn:
There are missing locations and population, I will go over them with you @hyperknot but it should not block this PR

hyperknot:
Can we split this PR to ones not using covid19-eu-data dataset and one doing only covid19-eu? That is a low quality dataset, uses cases for hospitalized in France, it has bad date format, like 2020-12-04T00:00 which makes no sense, like a buggy ISO date implementation.

We should have their datasets in a single directory and we'd give -1 as a priority, basically only using them for backfilling missing cache.

hyperknot:
CY is good
CZ is good
image

hyperknot:
EE is good
image

hyperknot:
NL is good
image

qgolsteyn:
Yes, this can be done

hyperknot:
- [ ] Can you make this PR only contain the ones I've commented on and move the rest to a different PR?
- [ ] For sources about potential scrapers, we should open an issue for each, not a directory in the scrapers dir, that should only contain real JS files.

hyperknot:
After that I'm happy to merge this.

qgolsteyn:
Removed all scrapers from `covid19-eu-data` and removed the TODO files

PR #774, opened 2020-04-13
DRAFT: Look for ArcGIS response and retry when fetching CSVs

shaperilio:
Requesting a CSV from an ArcGIS server could give a JSON response that tells us the server "is working on it". We previously ignored this; @jzohrab saw at least one instance of this response being cached.

This fixes that by retrying up to 5 times before giving up.

Note that in my testing with Panama, I had *several* problems getting the CSV, where it seemed the backend (on their side) was restarting the process often. In theory, instead of a fixed number of retries, we could give up once a certain amount of time was exceeded. 

Moreover, when I did get the CSV, the data was sometimes stale (i.e. it would not match what their dashboard was displaying) and we have no way of detecting that.

Anything that has an ArcGIS CSV almost certainly has a REST API to access it with, and we should transition all these to that. However, that requires handling pagination, which we are not currently doing (and would be the subject of another PR). I mention that here because this PR is then just a temporary fix until we (hopefully) transition to not pulling CSVs from ArcGIS sources anymore.

Refer to [this doc](https://docs.google.com/document/d/1__rE8LbiB0pK4qqG3vIbjbgnT8SrTx86A_KI7Xakgjk/edit#) if you have questions.

### DIFF
--- a/src/shared/lib/fetch/index.js
+++ b/src/shared/lib/fetch/index.js
@@ -86,26 +86,65 @@ export const jsonAndCookies = async (url, date, options = {}) => {
  */
 export const csv = async (url, date, options = {}) => {
   return new Promise(async (resolve, reject) => {
-    const resp = await get(url, 'csv', date, options);
+    // Get the source URL, and retry here only if it's an arcGIS response that the CSV isn't ready.
+    let arcGISTries = 0;
+    let resp = null;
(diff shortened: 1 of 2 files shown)


shaperilio:
Note to self: I need to test what happens if a file in a previous day's cache happens to contain this response; we wouldn't want it to go to the server at that point.

PR #777, opened 2020-04-13
Add tool compare reports

jzohrab:
## Summary

This adds a simple diffing script for generated CSV and JSON reports,
with human-friendly output.  This lets devs refactor and verify code
more safely.

The script usage is detailed in the docs.  In summary, here's how it's used:

Before starting work (or git checkout a SHA you want to use as
reference), generate baseline reports:

```
yarn start --date 2020-04-06 --onlyUseCache --location US/NV --writeTo zz_before
```

After doing your work, regenerate to a different folder using the same
cached data:

```
yarn start --date 2020-04-06 --onlyUseCache --location US/NV --writeTo zz_after
```

Use new script `tools/compare-report-dirs.js`, comparing the "left"
and "right" folders.  Each report is listed with the differences:

```
$ node tools/compare-report-dirs.js --left zz_before --right zz_after

data-2020-04-06.json
--------------------
- [3, Storey County, Nevada, United States]/deaths value: 0 != 41
- [5, Washoe County, Nevada, United States]/deaths value: 4 != -1
...
```

The interpretation of the diff is covered in the docs.

## Changes

- Added `--writeTo` command line option
- Sources are sorted, to simplify comparison
- New libs and tests

## Additional notes

I haven't done this yet, but this code could likely be used in some
form for automated regression testing and reporting.  I haven't done
that yet as it implies that we have some regression baseline data
committed, and I wasn't sure if we could do that without introducing
too many failing CI test runs.  This is still potentially a useful
tool for devs and QA.

### DIFF
--- a/docs/sources.md
+++ b/docs/sources.md
@@ -25,6 +25,7 @@ _Last updated: 2020-03-26_
   - [Testing sources](#testing-sources)
     - [Test coverage](#test-coverage)
     - [Manual testing](#manual-testing)
+    - [Manual regression testing](#manual-regression-testing)
 
 This guide provides information on the criterias we use to determine whether a source should be added to the project, and offer technical details
 regarding how a source can be implemented.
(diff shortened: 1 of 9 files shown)


jzohrab:
I did look for a while for such modules and there are some differs, but none that did what I wanted with the json, i.e. finding the nodes where things are different.  Maybe what I did is overkill, but ... I wanted to do it b/c it was v interesting (lame, I know!), and it's potentially useful for the future.  The CSV diffing was trivial.

So, I hear you, and I think it's still useful to get something like this in.  If it causes pain in terms of bugs or too much stupid maintenance, I'll remove it.  Sounds good?

PR #780, opened 2020-04-13
Panama scraper (exposes at province level)

shaperilio:
Granularity is actually quite good, but we don't have maps or population data for anything smaller than a province.

I updated the province-level population in Wikidata and @hyperknot will generate exports including those soon.

Thanks everyone for your help on this!

### DIFF
--- a/src/shared/lib/maintainers.js
+++ b/src/shared/lib/maintainers.js
@@ -66,6 +66,16 @@ const maintainers = {
     state: 'Seoul',
     city: 'Gwanak-gu',
     flag: '🇰🇷'
+  },
+  shaperilio: {
+    name: 'Emilio Graff',
+    email: '1@emil.io',
+    github: 'shaperilio',
+    country: 'USA',
+    state: 'CA',
+    county: 'Alameda County',
+    city: 'Oakland',
+    flag: '🇺🇸'
   }
 };
 
--- a/src/shared/scrapers/PA/index.js
+++ b/src/shared/scrapers/PA/index.js
@@ -139,7 +139,7 @@ const scraper = {
       note: 'Distritos were mapped into counties in this data set; data exists at corregimiento level.'
     }
   ],
-  // URLs were obtained by snopping. See
+  // URLs were obtained by snooping. See
   // https://docs.google.com/document/d/1__rE8LbiB0pK4qqG3vIbjbgnT8SrTx86A_KI7Xakgjk/edit#
 
   // List of "corregimientos", which we will consider counties.
--- a/src/shared/scrapers/PA/index.js
+++ b/src/shared/scrapers/PA/index.js
@@ -5,18 +5,16 @@ import maintainers from '../../lib/maintainers.js';
 import log from '../../lib/log.js';
 import provinceToIso2 from './province-to-iso2.json';
 
-function titleCase(string) {
-  let result = '';
-  for (let i = 0; i < string.length; i++) {
-    if (i === 0 || string.charAt(i - 1) === ' ') {
-      result += string.charAt(i).toUpperCase();
-    } else {
-      result += string.charAt(i).toLowerCase();
-    }
-  }
-  result = result.replace('Del', 'del'); // ugh.
-  return result;
-}
+/**
+ * @param {string} inputString
+ * @returns {string}
+ */
+const titleCase = inputString =>
+  inputString
+    .split(' ')
+    .map(([firstLetter, ...otherLetters]) => firstLetter.toUpperCase() + otherLetters.join('').toLowerCase())
+    .join(' ')
+    .replace('Del', 'del');
 
 const patientStatus = {
   FALLECIDO: 'deaths',
--- a/src/shared/scrapers/PA/index.js
+++ b/src/shared/scrapers/PA/index.js
@@ -95,7 +95,7 @@ function updateData(data, o) {
 function addEmptyStates(data) {
   // Instead of this, we could actually design the system to do this:
   // 1. Obtain a list of administrative levels entities at whatever granularity the disease data exist
-  // 2. Crate a template data result with all zeros
+  // 2. Create a template data result with all zeros
   // 3. Traverse through data and populate.
   for (const province in provinceToIso2) {
     let found = false;
--- a/src/shared/scrapers/PA/index.js
+++ b/src/shared/scrapers/PA/index.js
@@ -94,7 +94,7 @@ function updateData(data, o) {
 
 function addEmptyStates(data) {
   // Instead of this, we could actually design the system to do this:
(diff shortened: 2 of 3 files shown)


src/shared/lib/maintainers.js
```diff
@@ -66,6 +66,16 @@ const maintainers = {
     state: 'Seoul',
     city: 'Gwanak-gu',
     flag: '🇰🇷'
+  },
+  shaperilio: {
```
camjc, line 70: ⭐

src/shared/scrapers/PA/index.js
```diff
@@ -0,0 +1,299 @@
...
+  for (let i = 0; i < string.length; i++) {
+    if (i === 0 || string.charAt(i - 1) === ' ') {
+      result += string.charAt(i).toUpperCase();
+    } else {
+      result += string.charAt(i).toLowerCase();
+    }
+  }
+  result = result.replace('Del', 'del'); // ugh.
+  return result;
+}
```
camjc, line 19:
```suggestion
/**
 * @param {string} inputString
 * @returns {string}
 */
const titleCase = inputString =>
  inputString
    .split(' ')
    .map(([firstLetter, ...otherLetters]) => firstLetter.toUpperCase() + otherLetters.join('').toLowerCase())
    .join(' ')
    .replace('Del', 'del');
```

Also, is it worth moving the `replace Del del` to the calller instead of inside this titlecase function to keep it generic?

src/shared/scrapers/PA/index.js
```diff
@@ -0,0 +1,299 @@
...
+  // fail src/events/crawler/scrape-data/run-scraper.js:16:13
+  // county: null,
+  // city: null,
+  cases: 0,
+  deaths: 0,
+  quarantine: 0,
+  hospitalized: 0,
+  icu: 0,
+  recovered: 0
+};
```
camjc, line 53: Do we want to be returning explicit 0s when we actually have no idea what the value should be? Or is it better to not return a value in that case?

src/shared/scrapers/PA/index.js
```diff
@@ -0,0 +1,299 @@
...
+
+function getCountsFromLocation(location) {
+  const o = JSON.parse(JSON.stringify(templateDataEntry)); // "deep copy". Fuck.
```
camjc, line 56: Apparently in node you can do this now, haven't tried it yet but seems a lot nicer than going to a string and back.
```
import v8 from 'v8';

const cloneDeep = obj => v8.deserialize(v8.serialize(obj));
```

src/shared/scrapers/PA/index.js
```diff
@@ -0,0 +1,299 @@
...
+
+function updateData(data, o) {
+  let entryToUpdate = null;
+  for (const entry of data) {
+    if (entry.state !== o.state) {
+      continue;
+    }
+    if (entry.county !== o.county) {
+      continue;
+    }
```
camjc, line 74:
```suggestion
    if (entry.state !== o.state || entry.county !== o.county) {
      continue;
    }
```

src/shared/scrapers/PA/index.js
```diff
@@ -0,0 +1,299 @@
...
+    entryToUpdate.quarantine += newEntry.quarantine;
+    entryToUpdate.hospitalized += newEntry.hospitalized;
+    entryToUpdate.icu += newEntry.icu;
+    entryToUpdate.recovered += newEntry.recovered;
+  }
+}
+
+function addEmptyStates(data) {
+  // Instead of this, we could actually design the system to do this:
+  // 1. Obtain a list of administrative levels entities at whatever granularity the disesase data exist
```
camjc, line 97:
```suggestion
  // 1. Obtain a list of administrative levels entities at whatever granularity the disease data exist
```

src/shared/scrapers/PA/index.js
```diff
@@ -0,0 +1,299 @@
...
+  // 2. Crate a template data result with all zeros
```
camjc, line 98:
```suggestion
  // 2. Create a template data result with all zeros
```

src/shared/scrapers/PA/index.js
```diff
@@ -0,0 +1,299 @@
...
+  priority: 1,
+  country: 'iso1:PA',
+  sources: [
+    {
+      url: 'http://minsa.gob.pa/coronavirus-covid19',
+      name: 'Ministerio de Salud de la República de Panamá',
+      note: 'Distritos were mapped into counties in this data set; data exists at corregimiento level.'
+    }
+  ],
+  // URLs were obtained by snopping. See
```
camjc, line 142:
```suggestion
  // URLs were obtained by snooping. See
```

src/shared/scrapers/PA/index.js
```diff
@@ -0,0 +1,299 @@
...
+    //   Edad_inicial: '',
+    //   Rango_edades: '',
+    //   FID: '1',
+    //   FID2: '1'
+    // }
+
+    const data = [];
+    log(`Panama has ${caseList.length} cumulative cases.`);
+    this.url = this._caseListUrl; // required for source rating.
+    for (const c of caseList) {
```
camjc, line 265: possibly no need for all these abbreviations, `case` and `object` may be more immediately readable.

src/shared/scrapers/PA/index.js
shaperilio: Haha wow, what a change! I'm fine with this if you promise you ran it and it works. Regarding "del" - Title capitalization is predictably complicated, but more than that, who knows what capitalization anything is using anyway. Really what I'm doing here is "take all caps provinces and convert them to what my ISO2 list has"; so I'm fine moving the "del" conversion to `getProvinceIso2`.

src/shared/scrapers/PA/index.js
shaperilio: Good catch, it should be null. Will change.

src/shared/scrapers/PA/index.js
shaperilio: Cool, I'll try it.

src/shared/scrapers/PA/index.js
shaperilio: I think it would have to be ` ! (entry.state === o.state && entry.county === o.county)`? In any case, I'm not adding counties now because we don't have the geography for them, but I'll do something along these lines.

src/shared/scrapers/PA/index.js
shaperilio: Can't do this anyway because that template object is what makes this scraper return "zero" instead of "no data"

src/shared/scrapers/PA/index.js
shaperilio: Will change.

shaperilio:
Making significant changes to this; will open a new one.

PR #781, opened 2020-04-13
feat: Saudi Arabia scraper

qgolsteyn:
This PR adds a scraper for Saudi Arabia, provides data for cases, deaths, and recovered. It uses an official government source, and has been crosschecked with Worldometer for accuracy.


qgolsteyn:
image

qgolsteyn:
One region is missing from the official dataset. However, all numbers are correct nationally when aggregating.

src/shared/lib/fetch/index.js
```diff
@@ -17,7 +17,7 @@ const CHROME_AGENT =
 const DEFAULT_VIEWPORT = { width: 1280, height: 800, isMobile: false };
 
 const RESPONSE_TIMEOUT = 5000;
-const READ_TIMEOUT = 30000;
+const READ_TIMEOUT = 60000;
```
lazd, line 20: Big number ouch

PR #783, opened 2020-04-13
Fixing CA to one file

aed3:
## Summary

Removed old county files and updated the data source to be government file for CA. Unfortunately CA doesn't give tested or recovered data in the central source. 

The data is also always one day behind, so it grabs the data from the day before for the current date. (This would also fix NY's problem too).


lazd:
Can we keep the existing scrapers, add a completely new scraper for the government source, set the priority to `2`, and don't do the `date - 1` hack (instead, just take the most recent data if we request a date newer than what's there)?

We want to have multiple sources for the same data, even if the CA data is authoritative, and especially if it's a day behind.

cc @HarrietRowan

jzohrab:
I pinged @aed3 a couple of times on Slack; hopefully we can get this completed, if the work is still valid.  There are a ton of conflicts due to my changes, hopefully not too severe.  @lazd , thoughts on this?

jzohrab:
Lazd's suggestion is the best.  The new dataset is good, we should use that.  The existing _mercury-news one seems to now be pulling from the wrong sheet, it's stale, we should pull from timeseries sheet instead, if we're going to use it.

lazd:
`_mercury-news.js` is deprecated, `mercury-news.js` is the proper one

jzohrab:
Thanks @lazd , I forgot that _mercury-news.js would be ignored due to underscore.  I'm going to "officially" deprecate it with an error for ppl like me in the future.  Closing this, thanks.

PR #784, opened 2020-04-13
add additional assertion to ID, their site seems flaky

camjc:
## Summary
This scraper sometimes fails, I think they can't handle the traffic volume. Lets add an assertion early on in the pipe so we know if its the scraping logic or just the whole page that is (generally temporarily) broken.


camjc:
Haha, I just realised I added this to india not indonesia. Ah well can't hurt India either. Will add to Indonesia now.

lazd:
Not only can it not hurt, but it showed me that India was broken. Nice work!

PR #785, opened 2020-04-13
Fixed NY (Timeseries does not contain a sample for SCRAPE_DATE)

aed3:


src/shared/scrapers/US/NY/index.js
```diff
@@ -84,13 +84,16 @@ const scraper = {
...
 
     const dateField = 'Test Date';
 
     // FIXME when we roll out new TZ support!
-    const fallback = process.env.USE_ISO_DATETIME ? new Date(datetime.now.at('America/New_York')) : datetime.getDate();
+    const fallback = new Date(
+      process.env.USE_ISO_DATETIME ? datetime.today.at('America/New_York') : datetime.getDate()
+    );
     let scrapeDate = process.env.SCRAPE_DATE ? new Date(`${process.env.SCRAPE_DATE} 12:00:00`) : fallback;
+    scrapeDate.setDate(scrapeDate.getDate() - 1);
```
lazd, line 96: Confused as to why this is necessary, can you explain?

aed3: Because the most recent data is always a day behind, the old way makes it look 2 days in a row had the same data. This new way pushes all the data back by one day instead of just the most recent one.

aed3: image


See how the last two dates are the same

lazd: Mmm yeah the last two days are the same, but then tomorrow gets updated and it rises normally. Seems fine... @qgolsteyn @camjc @shaperilio thoughts?

camjc: Feels like ideally we’d not have a number when we don’t know it yet?

hyperknot: I think we'll have to always put the latest known number as the actual number. If they don't update over the weekend, then yes it'll be a flat part.

shaperilio: If the table has dates, then we know exactly the day the data represents. So we should not be manipulating anything, there should be no duplicating the last point because our scrape date is "in the future". There should simply be no data for April 12 if the table only has up to April 11.

It would then be the job of the front end (e.g. the map) to know that on April 12, it should show April 11 data (and notify the user on hover/click that this is the last update date). We are dealing with multiple governments un multiple timezones so no one is updating all at the same time.

src/shared/scrapers/US/NY/index.js
```diff
@@ -84,13 +84,16 @@ const scraper = {
   async scraper() {
     this.url = 'https://health.data.ny.gov/api/views/xdss-u53e/rows.csv?accessType=DOWNLOAD';
     this.type = 'csv';
-    const data = await fetch.csv(this.url, false);
+    const data = await fetch.csv(this.url, false, { alwaysRun: true });
```
shaperilio, line 87: Don't do this. I'm guessing you were having trouble getting fresh data, and that's because you have `date` set to `false`, which caches files in the `cache` folder without a date, so you will never get new data during testing unless you delete the `cache` folder. The `cache` folder does not exist in production.

jzohrab:
I pinged @aed3 a couple of times on Slack; hopefully we can get this completed, if the work is still valid.  With the recent changes I'm not sure it is anymore ... @hyperknot and @shaperilio, thoughts?

hyperknot:
How are we handling missing data for a given day? null values?

jzohrab:
> How are we handling missing data for a given day? null values?

@hyperknot - old question!  Missing data should be `undefined`.

hyperknot:
> @hyperknot - old question! Missing data should be `undefined`.

Thanks!

jzohrab:
Picking up work in another branch, PR https://github.com/covidatlas/coronadatascraper/pull/974 (still draft).  I'll address the conversation points here, though, and will close it when the draft is complete.

src/shared/scrapers/US/NY/index.js
jzohrab: Right, thanks @shaperilio .  Fixed in the new branch, `const data = await fetch.csv(this, this, this.url, 'default', false);`

src/shared/scrapers/US/NY/index.js
jzohrab: I agree with @shaperilio and @camjc - we mustn't make assumptions about the data if we don't have it.  Since we already have a standard for reporting missing data (use `undefined`), let's do that.

Am doing so in https://github.com/covidatlas/coronadatascraper/pull/974.

jzohrab:
Closing in favor of https://github.com/covidatlas/coronadatascraper/pull/974.

PR #787, opened 2020-04-13
fix: RU to iso codes

qgolsteyn:
This PR moves the RU scraper to use ISO codes. It also aggregates all data to the country level, which was not done previously.

@hyperknot: We are missing population data for Sevastopol, other than that it looks good!


qgolsteyn:
image

qgolsteyn:
Google Translate was my friend! Manually went through each one of them, I double-checked the mapping!

hyperknot:
Wow, this is rad! Thanks for putting in the effort to do this one!

hyperknot:
I've added missing populations.

PR #790, opened 2020-04-13
Use scraper for India

camjc:
- I broke (Sorry!), then fixed it.
- Larry shorted it to get the release out.
- Now let's use it again since it works.


lazd:
Wait there's a new data source as of 2020-4-12, you removed the cache-only scraper for it in this PR

PR #791, opened 2020-04-13
Add scraper for Thailand

camjc:
image

Apparently the first 'confirmed case' means 'Confirmed patients (people)' second 'confirmed case' means 'Patient confirms the group travels from abroad and is quarantined in a designated area of the state (person)' 🤷


camjc:
@shaperilio likes this, merging :)

PR #795, opened 2020-04-13
Panama scraper (redo)

shaperilio:
Panama scraper, delivering data at county level.

Observations from running `yarn timeseries -l PA -d 2020-04-13`:
 - Counties with the same name as a province appear to get assigned the lat/long of the province (@hyperknot) - see `data.csv`.
 - `timeseries.json` has no location info in it at all.
image

@camjc Thanks very much for your suggestions on the last pull request; as it turns out I was not returning the correct data and trying to get rid of the zeros ended up giving me NaNs which the system doesn't like, so this is basically a rewrite.


hyperknot:
Looks nice. Two Qs:
1. How does the returned countyLevel array look like? 
2. Can you add some basic tests, like the newer scrapers in Europe have?

shaperilio:
> Looks nice. Two Qs:
> 
> 1. How does the returned countyLevel array look like?
Here's the JSON and CSV output for `yarn start -l PA`

[yarn_start_PA.zip](https://github.com/covidatlas/coronadatascraper/files/4472650/yarn_start_PA.zip)

hyperknot:
> > Looks nice. Two Qs:
> > 
> > 1. How does the returned countyLevel array look like?
> 
> Here's the JSON and CSV output for `yarn start -l PA`
> 
> [yarn_start_PA.zip](https://github.com/covidatlas/coronadatascraper/files/4472650/yarn_start_PA.zip)

Thanks. So we should comment out the county code for now and only return the state one.

shaperilio:
Ah yes, the lat/long begin assigned to those counties messes up the map. Will comment out and file a bug.

shaperilio:
@hyperknot spot-check added in `test` subfolder.

lazd:
@shaperilio is this ready for review/merge?

shaperilio:
Yes indeed @lazd! Having a little snapshot idea as suggested by @hyperknot is great and I think it can be easily turned into a test but I don't want to hold this up

hyperknot:
Looks good, merging it!

image

PR #797, opened 2020-04-14
ArcGIS aware fetch

shaperilio:
Requesting a CSV from an ArcGIS server could give a JSON response that tells us the server "is working on it". We previously ignored this; @jzohrab saw at least one instance of this response being cached.

This fixes that by retrying up to 5 times before giving up.

Note that in my testing with Panama, I had *several* problems getting the CSV, where it seemed the backend (on their side) was restarting the process often. In theory, instead of a fixed number of retries, we could give up once a certain amount of time was exceeded. 

Moreover, when I did get the CSV, the data was sometimes stale (i.e. it would not match what their dashboard was displaying) and we have no way of detecting that.

Anything that has an ArcGIS CSV almost certainly has a REST API to access it with, and we should transition all these to that. However, that requires handling pagination, which we are not currently doing (and would be the subject of another PR). I mention that here because this PR is then just a temporary fix until we (hopefully) transition to not pulling CSVs from ArcGIS sources anymore.

Refer to [this doc](https://docs.google.com/document/d/1__rE8LbiB0pK4qqG3vIbjbgnT8SrTx86A_KI7Xakgjk/edit#) if you have questions.

Note this updated pull request includes a test.


shaperilio:
@ryanblock, calling attention to this also for li.

shaperilio:
Dear all, please merge or discard; this is ancient now. If this is fine then maybe I can help port it into li.

shaperilio:
Thanks to @jzohrab the cache reference is now fixed on this; @ryanblock @lazd do we merge this and then port it over to li?

camjc:
Japan is hitting the 10,000 limit in production btw. I'm keen on this 👍

jzohrab:
Closing (finally!) b/c we need to focus on pagination.  Another suggestion for this came up today, and I need to resolve this problem for Li anyway.  Cheers ... and sorry for the wait. :-(  Blarg!

PR #803, opened 2020-04-14
Use ISO codes for all states

hyperknot:
# Summary

Removed US state autodetection and made sure all of them return `iso2:US-CA` like code. This way all the normalization code could be simplified and later on a lot of code can be removed.

## Changes

Did a search & replace on all US files.
Manually fixed Covidtracking to use state codes.
Manually fixed NYT.

Deleted `US/_usa-cdc.js` I hope it's not a problem, basically it didn't provide any kind of information and was disabled before (underscore in the name).


hyperknot:
Tests are passing locally, but I'm puzzled by tests actually. `yarn test` says 380 out of 380 is passing.
However GitHub says:

```
  total:     566
  passing:   541
  failing:   25
  duration:  2m 51.1s
```

jzohrab:
LGTM, and chatted w/ Z. on Slack.  The current fails are due to our tests being flaky, sometimes.

PR #806, opened 2020-04-14
fix: parse.number should successfully parse floats

### LINKED ISSUE #805: parse.number(160.0) returns 1600
## Description

/title

## Steps to reproduce
1. `parse.number('160.0')`
2. See it parses as `1600`

## Expected behavior

It parses as `160`

## Additional context

Resulted in North Dakota going from 160 cases to 1600 because they added a float to the CSV formatting.

Yes, there are locales that consider `1.600` to be `1,600` (`1600`), but currently, I don't believe we're seeing that anywhere.

lazd:
## Summary

Parse floats as ints Instead of discarding the decimal, fixes #805


jzohrab:
Unrelated from this PR: We've had some cases where HTML tables are empty (content == ''), and this returns zero.  Should it return null?

Should there be a second method, parseInt?  Most uses of this method are for deaths, cases, etc.

lazd:
Yeah that's a tough one. Filed #807, let's discuss there.

lazd:
Gotta merge this to fix stuff for tonight's release

PR #809, opened 2020-04-14
Sample API change, and tests to verify.

jzohrab:
Sample proposed API call, and a set of unit tests to verify.  This will significantly simplify migrating the cache to the new layout.

All `fetch` methods will pass three args minimum: `fetch.<method>(ref_to_scraper, url, '<cache-key>' `

* `ref_to_scraper` is used so that the cache can get the scraper's file path name (via `_path`)
* `cache-key` will not be used in the current code, but will be useful for cache migration, b/c cache in v1.0 uses these to differentiate files.  In almost all cases, this string will be `'default'`.

The `fetch` method should also all be on one single line ... I know this is nitpicky, but it makes the stupid test much easier to write.

With the tests, the hack changes on US/PA pass for some lines, and fail with the following:

```
# /US/PA/index.js fetch coding conventions (line "fetch.page(this.url);")
ok 171 "fetch.page(this.url);" ends with ');'
not ok 172 Expected >=3 args to fetch (scraper, url, cacheKey), got 1

...

# /US/PA/index.js fetch coding conventions (line "fetch.page(this, this.url, 'default');")
ok 167 "fetch.page(this, this.url, 'default');" ends with ');'
ok 168 Expected >=3 args to fetch (scraper, url, cacheKey), got 3
ok 169 first arg is this or obj

...

# /US/PA/index.js fetch coding conventions (line "fetch.page(this.url,")
not ok 173 "fetch.page(this.url," ends with ');'
not ok 174 Expected >=3 args to fetch (scraper, url, cacheKey), got 2

```

I'll make all of these changes myself, and will also provide before/after runs for all cached files, verifying that the various report files don't change.

### DIFF
--- a/src/shared/scrapers/US/PA/index.js
+++ b/src/shared/scrapers/US/PA/index.js
@@ -90,7 +90,7 @@ const scraper = {
     '0': async function scraper() {
       this.url = 'https://www.health.pa.gov/topics/disease/Pages/Coronavirus.aspx';
       this.type = 'list';
-      const $ = await fetch.page(this.url);
+      const $ = await fetch.page(this, this.url, 'default');
       let counties = [];
       const $lis = $('li:contains("Counties impacted to date include")')
         .nextAll('ul')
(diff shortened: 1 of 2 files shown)


jzohrab:
I might be able to have `cache-key` be an optional argument, defaulting to "default".  It depends on how many scrapers supply other args, many just use the URL only.

lazd:
Hmm, I'd rather see us parsing the code instead of looking at it with a regex, that will get really brittle :-\

jzohrab:
Closing, thanks for looking!

PR #829, opened 2020-04-16
fix: AU QLD ever-changing HTML structure

camjc:
Why bother caring what the cells are made of, we just need to find the last row and iterate through the direct descendants.


qgolsteyn:
Could you include a test with this small change, so we can begin to check for any regression in historical should we need to update this scraper again? After that I'll be happy to merge!

PR #830, opened 2020-04-16
fix: deprecate failing US/CA county scrapers

qgolsteyn:
This PR deprecates currently failing US/CA county scrapers. I am proposing this change as I do not see us having the capacity to maintain these scrapers and we already have aggregate sources covering the state of California. I am still allowing the scraper to cache their URL in case we are able to revisit these sources at a later date.

### DIFF
--- a/src/shared/scrapers/US/CA/glenn-county.js
+++ b/src/shared/scrapers/US/CA/glenn-county.js
@@ -30,7 +30,7 @@ const scraper = {
       $('span:contains("Glenn County COVID-19 Cases")')
         .first()
         .text()
-        .match(/Cases: (\d+)/)[1]
+        .match(/Cases:.*(\d+)/)[1]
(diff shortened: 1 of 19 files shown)


src/shared/scrapers/US/CA/glenn-county.js
```diff
@@ -30,7 +30,7 @@ const scraper = {
       $('span:contains("Glenn County COVID-19 Cases")')
         .first()
         .text()
-        .match(/Cases: (\d+)/)[1]
+        .match(/Cases:.*(\d+)/)[1]
```
lazd, line 33: This one got fixed tho? yay!

qgolsteyn: Yup, this was a very small issue that I could fix easily

PR #831, opened 2020-04-16
new scraper: JP including prefecture-level

camjc:
## Summary
There is an ARCGIS for Japan, now we'll scrape it for case data at prefecture-level.

Would love any pointers from folks who have done a bit of ARCGIS before.

### DIFF
--- a/src/shared/scrapers/JP/index.js
+++ b/src/shared/scrapers/JP/index.js
@@ -19,8 +19,8 @@ const groupBy = (array, func) => {
   }, {});
 };
 
-const country = 'iso1:JP';
-const japanIso2Values = Object.values(iso2).filter(item => item.countrylevel_id.startsWith(country));
+const countryIso1 = 'JP';
+const japanIso2Values = Object.values(iso2).filter(item => item.iso2.startsWith(countryIso1));
 
 const prefectureSpecialCases = {
   Hokkaido: 'Hokkaidō'
@@ -29,11 +29,12 @@ const prefectureSpecialCases = {
 const getIsoFromPrefectureName = prefectureName => {
   const modifiedName = prefectureSpecialCases[prefectureName] || prefectureName;
   const foundItem = japanIso2Values.find(({ name }) => name.startsWith(modifiedName));
+  assert(foundItem, `no item found for ${prefectureName}`);
   return foundItem.countrylevel_id;
 };
 
 const scraper = {
-  country,
+  country: `iso1:${countryIso1}`,
   maintainers: [maintainers.camjc],
   priority: 1,
   sources: [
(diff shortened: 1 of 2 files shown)


hyperknot:
Wow, everything found? Can you include a map screenshot? (that's the only way to check if a new scraper covers all regions correctly).

hyperknot:
Thanks @shaperilio for the map screenshot, it looks perfect to me.

shaperilio:
Map looks good:
image

@camjc the data source seems to be a list of patients, so you just count them by prefecture to establish number of cases. There are some interesting fields in there:
image

You can see `Sum_` is already the total cases. There's also a `SumByStatus` and `Status`, so I bet there's a lot of info here - and maybe the sum total actually also includes recovered people. I think we should merge this ASAP to start collecting data, but please open an issue to get these translated so we can see if there's more to parse here.

Regarding ArcGIS JSONs:
Not sure what `pjson` is for the format, but my browser's JSON viewing plugin doesn't trigger with that, when it does with `json`. Data seems to be the same.

FYI you can change that to `html` in a browser and get a [query editor](https://services8.arcgis.com/JdxivnCyd1rvJTrY/arcgis/rest/services/v2_covid19_list_csv/FeatureServer/0/query?where=0%3D0&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&resultType=none&distance=0.0&units=esriSRUnit_Meter&returnGeodetic=false&outFields=*&returnGeometry=false&featureEncoding=esriDefault&multipatchOption=xyFootprint&maxAllowableOffset=&geometryPrecision=&outSR=&datumTransformation=&applyVCSProjection=false&returnIdsOnly=false&returnUniqueIdsOnly=false&returnCountOnly=false&returnExtentOnly=false&returnQueryGeometry=false&returnDistinctValues=false&cacheHint=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&having=&resultOffset=&resultRecordCount=&returnZ=false&returnM=false&returnExceededLimitFeatures=true&quantizationParameters=&sqlFormat=none&f=html). 

It seems the query fields that are available depend on some configuration, because this is more than I've seen. The minimum URL required to get this data is:

https://services8.arcgis.com/JdxivnCyd1rvJTrY/arcgis/rest/services/v2_covid19_list_csv/FeatureServer/0/query?where=0%3D0&outFields=*&f=json

# IMPORTANT
The REST API for arcGIS requires pagination, see #832. Japan will hit its limit in the next few days (if not tomorrow). So I vote to merge this ASAP but work on #832 urgently.

camjc:
Sweet, we can run the groupBy on the status field and return the length of each status once we figure out what they mean. Seems like the statuses are based on what week vs today they were identified? So if we filter out by date (to make timeseries) then the statuses may not be relevant. Unsure since google translate isn’t enlightening in this case.

shaperilio:
I made an issue to translate. #833

shaperilio:
@camjc if you're interested, you can navigate up their ArcGIS tree and see what else they have. I found out that Panama has a country-level time series along with the patient list.

camjc:
Looks the same as what’s going through this scraper? Do some logging through before or after grouping by prefecture.

Issue #832, opened 2020-04-16
Pulling from ArcGIS REST API requires supporting pagination


shaperilio:
## Description
There are two ways to get data from ArcGIS sources: snooping around for the "opendata" CSV, and using the source's REST API. The REST API is preferred in my opinion; you sometimes have to wait for the CSV to be generated (see [PR](https://github.com/covidatlas/coronadatascraper/pull/797)) and I've gotten stale data before. See [doc[(https://docs.google.com/document/d/1__rE8LbiB0pK4qqG3vIbjbgnT8SrTx86A_KI7Xakgjk/edit)

Say that you find an ArcGIS REST API URL, e.g. this one for Japan:
https://services8.arcgis.com/JdxivnCyd1rvJTrY/arcgis/rest/services/v2_covid19_list_csv/FeatureServer/0/query?where=0%3D0&outFields=*&f=json

If you go to [everything up to that last "0"](https://services8.arcgis.com/JdxivnCyd1rvJTrY/arcgis/rest/services/v2_covid19_list_csv/FeatureServer/0), you get a description with an important configuration parameter, "Max Record Count". Japan has this at 10k, and returns a list of cases, so if the total number of cases in japan ever exceeds 10k, we won't see the count increase.

Pagination is handled with two query fields: `resultOffset` and `resultRecordCount` (you can see a query GUI if you change `f=json` to `f=html` in the URL).

If your query exceeds the max record count, the resulting JSON will have a field `"exceededTransferLimit": true` in it.

## What we need to do

We should have a `fetch.arcGISjson(featureLayerURL, ...)` which is aware of this. `featureLayerURL` should be everything up to that last zero, e.g.

https://services8.arcgis.com/JdxivnCyd1rvJTrY/arcgis/rest/services/v2_covid19_list_csv/FeatureServer/0

To that, we need to add the minimum query statement, which I _believe_ is:

`where=0%3D0&outFields=*&f=json` (`%3D` is `=`)

*NOTE*: it's probably a good idea to add `returnExceededLimitFeatures=true` to the query; it seems on by default, I can't imagine why anyone would not want to do that and still give us data, and I can't find documentation on it.

The fetcher should check if the response has `"exceededTransferLimit": true` in it. If yes, then query again, with 

`where=0%3D0&outFields=*&resultOffset=n&f=json` 

where `n` is equal to the length of the `features` key in the JSON result. (I don't think we need the `resultRecordCount` parameter in there).

[...]

PR #835, opened 2020-04-16
feat: add ArcGIS REST API fetch function

qgolsteyn:
This PR creates a new function to fetch a JSON from an ArcGIS dashboard's REST API. It follows the suggestion from #832 by handling pagination, and the approach is compatible. I have update the SE scraper to demonstrate intended behavior. Any non-timeseries scraper will require a more significant change to avoid creating a cache-miss due to the changed URL.


src/shared/lib/fetch/index.js
```diff
@@ -270,8 +270,44 @@ export const getArcGISCSVURLFromOrgId = async function(serverNumber, orgId, laye
...
+export const getArcGISCSVURL = async (serverNumber, dashboardId, layerName) => {
   const dashboardManifest = await json(`https://maps.arcgis.com/sharing/rest/content/items/${dashboardId}?f=json`);
   const { orgId } = dashboardManifest;
   return getArcGISCSVURLFromOrgId(serverNumber, orgId, layerName);
 };
+
+/**
+ * Retrieves data from an ArcGIS REST API. By default, it will retrieve all items at the provided linked with geometry turned off.
+ * You can control pagination size through the `k` parameter in `options`
+ * @param {string} url URL of the resource
```
lazd, line 282:
```suggestion
 * @param {string} featureLayerURL URL of the resource
```

src/shared/lib/fetch/index.js
```diff
@@ -270,8 +270,44 @@ export const getArcGISCSVURLFromOrgId = async function(serverNumber, orgId, laye
...
+ * @param {*} date the date associated with this resource, or false if a timeseries data
```
lazd, line 283: Does this work with various and sundry data sources that store dates in different formats?

src/shared/lib/fetch/index.js
```diff
@@ -270,8 +270,44 @@ export const getArcGISCSVURLFromOrgId = async function(serverNumber, orgId, laye
...
+ * @param {object} options customizable options:
+ *  - k: number of features we want to receive for each request. A smaller number means more request to grab the complete dataset,
+ *  a larger number may result in a partial dataset if we request more than `Max Record Count`. Defaults to 500.
+ *  - additionalParams: additional parameters for this request. Defaults to `where=0%3D0&outFields=*&returnGeometry=false`.
+ *  - alwaysRun: fetches from URL even if resource is in cache, defaults to false
+ *  - disableSSL: disables SSL verification for this resource, should be avoided
+ */
+export const arcGISjson = async (featureLayerURL, date, options = {}) => {
```
lazd, line 291:
```suggestion
export const arcGISJSON = async (featureLayerURL, date, options = {}) => {
```

src/shared/lib/fetch/index.js
```diff
@@ -270,8 +270,44 @@ export const getArcGISCSVURLFromOrgId = async function(serverNumber, orgId, laye
...
+  const { k, additionalParams } = {
```
lazd, line 292: prefer descriptive var names instead of single chars

src/shared/lib/fetch/index.js
qgolsteyn: All scrapers currently accessing ArcGIS json data will see a bunch of cache miss as the URL will change. So we will have to take into account. Regarding the source date format, this should not be a problem. The date field is for caching only (like the other fetch functions). The format of dates in the json itself does not influence how this function works.

shaperilio: If we switch a scraper from CSV to this then shouldn't we just make a new scraper in the array with the date of the switch?

src/shared/lib/fetch/index.js
```diff
@@ -270,8 +270,44 @@ export const getArcGISCSVURLFromOrgId = async function(serverNumber, orgId, laye
...
+    additionalParams: 'where=0%3D0&outFields=*&returnGeometry=false',
+    ...options
+  };
+
+  const url = `${featureLayerURL.replace(/\?.*$/, '')}?f=json${additionalParams ? `&${additionalParams}` : ''}`;
+
+  const output = [];
+
+  let n = 0;
+  let response = await json(`${url}&resultOffset=${n}&resultRecordCount=${n + k}`, date, options);
```
shaperilio, line 303: @qgolsteyn apologies, but I had a typo in my original issue description (which I've now corrected). 

```suggestion
  let response = await json(`${url}&resultOffset=${n}&resultRecordCount=${k}`, date, options);
```

src/shared/lib/fetch/index.js
```diff
@@ -270,8 +270,44 @@ export const getArcGISCSVURLFromOrgId = async function(serverNumber, orgId, laye
...
+
+  while (response && response.features.length > 0) {
+    n += k;
+    output.push(...response.features.map(({ attributes }) => attributes));
+
+    response = await json(`${url}&resultOffset=${n}&resultRecordCount=${n + k}`, date, options);
```
shaperilio, line 309:
```suggestion
    response = await json(`${url}&resultOffset=${n}&resultRecordCount=${k}`, date, options);
```

src/shared/lib/fetch/index.js
```diff
@@ -270,8 +270,44 @@ export const getArcGISCSVURLFromOrgId = async function(serverNumber, orgId, laye
...
+  };
+
+  const url = `${featureLayerURL.replace(/\?.*$/, '')}?f=json${additionalParams ? `&${additionalParams}` : ''}`;
+
+  const output = [];
+
+  let n = 0;
+  let response = await json(`${url}&resultOffset=${n}&resultRecordCount=${n + k}`, date, options);
+
+  while (response && response.features.length > 0) {
```
shaperilio, line 305: I think you need this:

```suggestion
  while (response && response.features && response.features.length > 0) {
```

src/shared/lib/fetch/index.js
```diff
@@ -270,8 +270,44 @@ export const getArcGISCSVURLFromOrgId = async function(serverNumber, orgId, laye
...
+    n += k;
```
shaperilio, line 306:
```suggestion
    n += response.features.length;
```

So that if our choice of `k` exceeds their options, we don't skip records.

shaperilio: @jzohrab regarding `cachekey`: 

Let's say we have a source with 603 cases. We initially set `k=1000`. And let's say the source has limited queries to return a maximum of 500 records.
First fetch will be `resultOffset = 0; resultRecordCount = 1000`, and receive 500 records.
Second fetch will be `resultOffset = 500; resultRecordCount = 1000`, and receive 103 records.
Third fetch will be `resultOffset = 603; resultRecordCount = 1000`, and receive 0 records.
So on that day, we will cache three files, provided that `cachekey` contains `resultOffset`'s value in it.

Let's say at the next scrape date, there are 1203 cases. You can see immediately that `resultOffset` will be different for the third fetch, and thus its `cachekey` would change. BUT, this is a new scrape date, so it should be fetching everything from the server anyway, right?

jzohrab: @shaperilio - yes, that's right.  New scrape date won't hit the cache.

This method will work with the caching in the current system, b/c the URL is encoded to a hash.

Doesn't the result set contain something like a "nextPageToken" or something for the next request?  It looks like this just keeps fetching with offsets and recordcounts.  That will also work with caching, no worries. :-)

jzohrab:
This will need some work to function with the arcgis stuff currently in master. :-)

src/shared/lib/fetch/index.js
qgolsteyn: Yup

qgolsteyn:
Thank you for all these comments, second round of review is now open :)

src/shared/lib/fetch/index.js
camjc: Theres an eslint rule for this if we want to enable it.

src/shared/lib/fetch/index.js
```diff
@@ -292,3 +292,52 @@ export const getArcGISCSVURL = async function(scraper, serverNumber, dashboardId
...
 };
+
+/**
+ * Retrieves data from an ArcGIS REST API. By default, it will retrieve all items at the provided linked with geometry turned off.
+ * You can control pagination size through the `k` parameter in `options`
+ * @param {*} scraper the scraper object
+ * @param {string} featureLayerURL URL of the resource
+ * @param {*} date the date associated with this resource, or false if a timeseries data
+ * @param {object} options customizable options:
+ *  - k: number of features we want to receive for each request. A smaller number means more request to grab the complete dataset,
```
shaperilio, line 303:
```suggestion
 *  - featuresToFetch: number of features we want to receive for each request. A smaller number means more request to grab the complete dataset,
```

src/shared/lib/fetch/index.js
shaperilio: Oh I understand what you mean; anything previously using the REST API will have cache misses if we switch them all to this new call...

shaperilio:
I put a PR into @qgolsteyn's branch that migrates all the scrapers that would be affected.

src/shared/lib/fetch/index.js
shaperilio: Fixed.

src/shared/lib/fetch/index.js
shaperilio: All scrapers were migrated.

shaperilio:
@qgolsteyn @jzohrab @camjc @lazd pinging you folks on this PR.

@jzohrab , I believe you wanted @qgolsteyn to remove a submodule update?

Note:
I migrated every scraper that this would affect in a potentially ugly way, but so that we don't get cache misses due to different URLs. Part of this ugliness is this addition to `datetime`:

```
const ARCGIS_PAGINATION_DEPLOY_DATE = '2020-04-19'
```

We need to set that to the date we want these sources to start using the paginated fetch. So it has to at least be the day that we merge. If scrape date is before this date, the scrapers will use their old URLs. I believe this can all go away with `li`'s cache architecture?

Testing:
I didn't write any tests. However, for each scraper affected, I ran a timeseries across this date boundary to ensure caching and fetching in the new way worked. This was an eyeball test, quickly checking e.g. that cases didn't plummet to zero.

shaperilio:
See #889

jzohrab:
Closing this PR, as the work is part of #889.  Thank you @qgolsteyn !

PR #837, opened 2020-04-16
Convert PA to a timeseries

shaperilio:
Thanks to @qgolsteyn and @camjc's idea, we can exploit the patient list we get to make a time series.

Note: also converted to fetching JSON because we've been getting stale CSV data.

Known issues / hacks:
 - Does its own ArcGIS pagination in case we couldn't merge @qgolsteyn PR for official support of that.
 - The two most recent data points in the timeseries are equal, which I think may be a timezone thing that will go away in li.


qgolsteyn:
Why merge this now? Weren't we waiting on my PR?

PR #841, opened 2020-04-17
actually deprecate US CA Santa Clara county

camjc:
##  Summary
Looks like we accidentally chucked the wrong month in the date. 🤷


camjc:
This is failing the test suite, isn't this how we should be handling deprecations?

shaperilio:
> This is failing the test suite, isn't this how we should be handling deprecations?

I think that is OK. @jzohrab ?

PR #844, opened 2020-04-17
fix US MI scraper, still seems slightly jank

camjc:
## Summary
caption changed, so reduced the words to make it match both cases.

## Changes
While I was there I overhauled the heading matching so its not as brittle.
I still dislike how this is doing the monkey-patching around Detroit, hopefully someone can clean that up, and theres some weirdness about it throwing out certain rows that I don't understand.

### DIFF
--- a/src/shared/scrapers/US/MI/index.js
+++ b/src/shared/scrapers/US/MI/index.js
@@ -1,10 +1,32 @@
+import assert from 'assert';
 import * as fetch from '../../../lib/fetch/index.js';
 import * as parse from '../../../lib/parse.js';
 import * as transform from '../../../lib/transform.js';
 import * as geography from '../../../lib/geography/index.js';
+import getKey from '../../../utils/get-key.js';
 
 // Set county to this if you only have state data, but this isn't the entire state
 const UNASSIGNED = '(unassigned)';
+const DetroitCity = 'Detroit City';
+const WayneCounty = 'Wayne County';
+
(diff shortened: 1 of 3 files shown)


src/shared/scrapers/US/MI/index.js
```diff
@@ -125,55 +154,62 @@ const scraper = {
...
-      if (county === 'Out of State' || county === 'Other' || county === 'Not Reported' || county === 'Unknown') {
-        unassignedObj.cases += cases;
-        unassignedObj.deaths += deaths;
+      if (['Out of State', 'Other', 'Not Reported', 'Unknown'].includes(rowData.county)) {
+        unassignedObj.cases += rowData.cases;
+        unassignedObj.deaths += rowData.deaths;
         return;
       }
 
+      // TODO: What is this doing and is it legitimate?
```
shaperilio, line 187: I wonder if there used to be a total row at the end?

src/shared/scrapers/US/MI/index.js
```diff
@@ -125,55 +154,62 @@ const scraper = {
...
       if (index < 1 || index > $trs.get().length - 2) {
         return;
```
shaperilio, line 189: Is this not just to handle the header row and the totals row at the end?

camjc: Yeah but seems to me like it’s ignoring the first county, and normally we exclude via css

PR #845, opened 2020-04-17
Update country-levels to v2.0.0

hyperknot:
New major version using a much better GeoJSON simplification. 

features.json down from 50 MB to 15 MB!

By converting it to TopoJSON later on, we can make it to 5 MB, which is only 1.7 MB gzipped (all countries, states and counties together!)

timeseries.csv is the same as in master, except for some lower digit differences in center coordinates.
map is visually the same as in master.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "ajv": "^6.12.0",
     "cheerio": "^1.0.0-rc.3",
     "cheerio-tableparser": "^1.0.1",
-    "country-levels": "https://github.com/hyperknot/country-levels/releases/download/v1.8.0/export_q7.tgz",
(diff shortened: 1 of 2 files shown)


lazd:
How much different does it look with the simplified GeoJSON? Got some examples?

lazd:
Any reason not to go straight to TopoJSON? Looks like mapbox can read it just fine.

hyperknot:
> How much different does it look with the simplified GeoJSON? Got some examples?

It should look the same, if not better. It's mostly smaller because it removed 330,000 little islands scattered around Canada, etc.

For publishing the data we should still publish GeoJSON. For the map, yes we should output TopoJSON, but it's independent from this PR.

hyperknot:
new:
new

old:
old

hyperknot:
Actually, it's definitely more detailed in v2.0:
new:
new

old:
old

PR #900, opened 2020-04-26
Add dumpRaw for regression testing with li.

jzohrab:
## Summary

This dumps raw scraped data to a new report, `raw.json`, stored in dist.

Sample output: 

```
[
  {
    "_path": "/Users/jeff/Documents/Projects/coronadatascraper/src/shared/scrapers/US/CA/san-diego-county.js",
    "county": "fips:06073",
    "state": "iso2:US-CA",
    "country": "iso1:US",
    "timeseries": false,
    "certValidation": true,
    "priority": 0,
    "cases": 2943,
    "deaths": 111
  }
]
```

The new routine only includes out a handfull of fields from the raw scraper data in the report, and the resulting file is ~ 3 mb.

With this output, we can work on getting regression testing in place for corresponding sources in li.

## Changes

* added `--dumpRaw` to options, eg `yarn start --dumpRaw`.  The raw report isn't written if that's not set.

### DIFF
--- a/src/events/processor/write-data/dump-regression-raw.js
+++ b/src/events/processor/write-data/dump-regression-raw.js
@@ -2,6 +2,8 @@ import path from 'path';
 import * as fs from '../../../shared/lib/fs.js';
 
 const writeRawRegression = async args => {
+  if (!args.options.dumpRaw) return args;
+
   let suffix = '';
   if (args.options.outputSuffix !== undefined) {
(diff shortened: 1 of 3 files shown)


jzohrab:
Pretty straightforward, biggest thing to note is that if `--dumpRaw` isn't specified, it doesn't print, so no affect on existing code.  `if (!args.options.dumpRaw) return args;`

PR #902, opened 2020-04-27
Fix vt cache key

jzohrab:
And add a migration check.


jzohrab:
Merging this, it doesn't touch existing production code.

Issue #928, opened 2020-05-01
US/CO scraper broken


jzohrab:
`yarn start --location US/CO`

  🐢  Cache miss for scraper: /US/CO/index.js; url: https://opendata.arcgis.com/datasets/fbae539746324ca69ff34f086286845b_0.csv; filepath: coronadatascraper-cache/2020-5-1/2dadf4f54d967d4e31058861ac745674.csv
  🚦  Loading data for https://opendata.arcgis.com/datasets/fbae539746324ca69ff34f086286845b_0.csv from server
  ❌ Got error 404 trying to fetch https://opendata.arcgis.com/datasets/fbae539746324ca69ff34f086286845b_0.csv
  ❌ Error processing iso2:US-CO, iso1:US:  TypeError: data is not iterable
    at Object.2020-03-20 (/Users/jeff/Documents/Projects/coronadatascraper/src/shared/scrapers/US/CO/index.js:131:28)

jzohrab:
URL in the scraper gives: `{"errors":[{"title":"Resource Not Found","status":404,"message":"Not Found"}],"meta":{}}`

jzohrab:
They have an HTML table at https://covid19.colorado.gov/covid-19-data.

lazd:
Perhaps this explains #930

Issue #930, opened 2020-05-01
CO-US showing 0 cases from 2020-4-23 to 2020-4-27


lazd:
## Location, date, and short issue description
https://staging.covidatlas.com/colorado-united-states

## Source

US-CO

## Snippet/screenshot

![image](https://user-images.githubusercontent.com/201344/80826385-f7fb0580-8b96-11ea-84e6-95c271beaa94.png)

jzohrab:
With the fix, we're now like this:

![image](https://user-images.githubusercontent.com/1637133/80869239-627e7500-8c6d-11ea-941a-93dd7ddbbb3c.png)

While it's not non-zero, it's the best we can do, unless we back-fill, but we have bigger things to worry about. :-)

PR #931, opened 2020-05-01
Fix US/CO.

jzohrab:
Resolves https://github.com/covidatlas/coronadatascraper/issues/928
 
`yarn start --location US/CO` works for today.  data.json extract:

```
{
    "state": "Colorado",
    "country": "United States",
    "aggregate": "county",
    "url": "https://opendata.arcgis.com/datasets/7dd9bfa5607f4c70b2a7c9634ccdca53_0.csv",
    "cases": 15284,
    "deaths": 777,
    "hospitalized": 2697,
    "tested": 72390,
    "rating": 0.49019607843137253,
    "coordinates": [
      -105.551,
      38.9975
    ],
    "tz": [
      "America/Denver"
    ],
    "featureId": "iso2:US-CO",
    "population": 5758736,
    "populationDensity": 21.330560753945722,
    "countryId": "iso1:US",
    "stateId": "iso2:US-CO",
    "name": "Colorado, United States",
    "level": "state"
  }
```

```
    "cases": 15284,
    "deaths": 777,
```

Match the county numbers rolled up, and hosp/tested/pop at the state level are read from the state record.

Added some assertions to ensure that the format is actually valid. :-P

### DIFF
--- a/src/shared/scrapers/US/CO/index.js
+++ b/src/shared/scrapers/US/CO/index.js
@@ -4,6 +4,8 @@ import * as transform from '../../../lib/transform.js';
 import * as rules from '../../../lib/rules.js';
 import * as geography from '../../../lib/geography/index.js';
 
+const assert = require('assert');
+
 // Set county to this if you only have state data, but this isn't the entire state
 const UNASSIGNED = '(unassigned)';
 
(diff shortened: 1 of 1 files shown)


lazd:
Does this fix #930 too?

jzohrab:
> #930

I doubt it, I changed the URL.  If we don't have that in the cache, we're out of luck.  I'll try running one of those dates.

jzohrab:
Ah hang on, that URL in the cache is still there, at least for 2020-04-25.  I'll see if I can sort that out too.

jzohrab:
I added a "fix" that at least parses the cached data for those days ... but the cache is stale (flat numbers).  But it's not zero for Apr 23-27.

jzohrab:
I actually left the console log in on purpose, b/c the data was whack.  :-)  But we can remove it, it won't mean anything.  Re the "|| 0", I'm not sure!

PR #934, opened 2020-05-01
updated typography

NateBaldwinDesign:
## Summary
Stylistic updates for typography & map visualization colors

Map vis colors changed to use a palette that is industry standard due to the similarity of the two palettes. Some of the lower numbers on the map may not be as noticeable, however the appropriate way to do this would be to apply logarithmic scale to the data prior to mapping it to a color scale (rather than the current modified color scale that approximates this transformation)

## Changes
- Heading font family and weight
- Body text font family
- Data vis now using standard Magma color palette

## Additional notes
image
image


lazd:
This bad boy needs a rebase. I'll handle it, if it's still a relevant change you want to get in, @NateBaldwinDesign?

jzohrab:
I just tried to merge master into this and got a ton of conflicts.  If there is anything to be salvaged from this, it should be done soon!

I pinged @NateBaldwinDesign on Slack, no response.  I'll close this in a few days unless asked to leave it open.

jzohrab:
Discussed with Nate on Slack.  This is too stale to sort out, so he'll start a new branch with relevant parts.  Cheers! z

PR #951, opened 2020-05-03
Scrape to raw gen from raw

jzohrab:
## Summary

This breaks apart the implicit argument passing of src/shared/index.js generate, so that we know what's getting used where.

As a side effect, this allows for scrapes to generate raw data files (in dist-raw), and gives the option to generate the reports using these raw data files.  These are two new cli methods, `raw:scrape` and `raw:report`, so that we don't get confused with these and the existing `yarn start`.

The diffs look scattered ... the individual commits might be easier to read, to understand the progression of the change.

## Changes

* Major change: breaks apart the implicit argument passing of src/shared/index.js generate, so that we know what's getting used where.
* Changed the function calls to standardize how the "report summary" is built
* Add `raw:scrape` and `raw:report` yarn scripts and js files
* Changed raw data output to include all necessary parts so that reports generated from dist-raw have parity with reports generated the usual way.

## Testing done

I regression tested this by generating reports off of master to dir `zz-master` using cache-only (and turning off my wifi to be sure, because sometimes that skewed results):

```
yarn start --date 2020-04-29 -x --writeTo zz-master
```

and then generating reports using this branch (saving to raw files, and then generating from raw files) to dir `zz-broken-chain`:

```
yarn raw:scrape --date 2020-04-29 -x
yarn raw:report --date 2020-04-29 -x --writeTo zz-broken-chain
```

Then I checked with the regression tool `node tools/compare-report-dirs.js --left zz-master --right zz-broken-chain/`, and all were equal.

I've only checked the above date, but can cycle through all of them if needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ public/*.js.map
 coronavirus-data-sources/
 dist/
 dist-raw/
+dist-raw-li/
 node_modules
 public/*.json
 sam.json
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ public/*.js
 public/*.js.map
 coronavirus-data-sources/
 dist/
+dist-raw/
 node_modules
 public/*.json
 sam.json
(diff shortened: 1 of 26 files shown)


jzohrab:
Closing in favor of https://github.com/covidatlas/coronadatascraper/pull/968.

PR #954, opened 2020-05-04
Fix NL when data not available.

jzohrab:
If data is not available, old code returned empty array, which then caused
validation error 'iso1:NL: Error: Invalid data: scraper for iso1:NL returned 0 rows'.
Returning undefined follows our convention for 'empty' data, and doesn't
raise an error.

I'm not 100% sure this is the right thing to do, but it at least stops the error being thrown.

### DIFF
--- a/src/shared/scrapers/NL/index.js
+++ b/src/shared/scrapers/NL/index.js
@@ -31,7 +31,6 @@ const scraper = {
     const nationalUrl =
       'https://raw.githubusercontent.com/J535D165/CoronaWatchNL/master/data/rivm_NL_covid19_national.csv';
     const nationalData = await fetch.csv(this, nationalUrl, 'national', false);
-
     const hospitalized = nationalData.find(
       item => datetime.scrapeDateIs(item.Datum) && item.Type === 'Ziekenhuisopname'
     );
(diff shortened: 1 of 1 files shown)


lazd:
I'd rather return data from yesterday than nothing... Can we do that, like the rest of the timeseries scrapers (NYT, etc)?

jzohrab:
I didn't see this comment!  I just submitted a PR that changes NYT to return 'undefined' if we don't have data. :-)

I feel we _shouldn't_ return data from yesterday, that feels like an adjustment we can make during reporting/front-end.  We should sort that out, that's a big difference, but I think it's the right thing to do (to differentiate between real data we actually have vs inferences we're making).

jzohrab:
Yes, thinking more on this, I def. feel this is not the right thing to do.  It's easy to write checks that deal with data that we know is real, vs data that we're making up.

jzohrab:
Updated so that NL returns yesterday's data when today's isn't ready.  That seems ok to me, given how flexible time is, and it's not terrible to report yesterday's.  One potential issue: if the source goes stale, we'll always report the same thing ... but that's a common problem anyway.  Cheers @lazd , merging this one, it seems OK to me and passes locally.

PR #974, opened 2020-05-07
Fix US/NY "Timeseries does not contain a sample for SCRAPE_DATE" error

jzohrab:
Replacement for https://github.com/covidatlas/coronadatascraper/pull/785.  Boils down to:

```
    if (counties.length === 0) {
      console.log(`No data for scrape date ${scrapeDate}, returning undefined.`);
      return [{ cases: undefined, tested: undefined }];
    }
```

### DIFF
--- a/src/shared/scrapers/US/NY/index.js
+++ b/src/shared/scrapers/US/NY/index.js
@@ -84,51 +84,43 @@ const scraper = {
   async scraper() {
     this.url = 'https://health.data.ny.gov/api/views/xdss-u53e/rows.csv?accessType=DOWNLOAD';
     this.type = 'csv';
+
     const data = await fetch.csv(this, this.url, 'default', false);
 
-    const dateField = 'Test Date';
-
     // FIXME when we roll out new TZ support!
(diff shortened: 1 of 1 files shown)


lazd:
Why would we return anything at all here? What purpose does that serve? Throwing seems to be a fine solution and is done in many other places.

jzohrab:
> Why would we return anything at all here? What purpose does that serve?

I actually wasn't sure what the best solution was, within the system model.  I returned "undefined" because that's what the docs say we should return when there isn't any data (for a particular field).

If we return null or undefined, an exception occurs further up the stack.

I didn't feel that throwing was the right thing to do, for a two reasons.  Main reason: the error is actually getting reported as an error in Slack, and will be in any other monitoring we add.  Secondary: if it's happening all the time, it's not really an exception, it's a property of the data.

I didn't see any place throwing errors as a matter of processing choice, except for "DeprecatedError".  The other errors in #scraper-status in Slack don't seem to be equivalent to this error.

With all that, this is where I ended up.  It felt like the correct thing to do.

jzohrab:
I have handled this differently in Li : the "sources/_lib/timeseries-filter.js" returns the latest filter date possible if needed, and the records can be given that filter date explicitly.  This is no longer needed.  Closing! z

PR #991, opened 2020-05-11
feat: add analytics

lazd:


lazd:
perms added for @ryanblock @jzohrab and Andi

PR #968, opened 2020-05-06
Generate reports with cds and li

jzohrab:
This is a companion PR for https://github.com/covidatlas/li/pull/89, to generate reports using data from both CDS and Li.

## Summary

This breaks apart the implicit argument passing of src/shared/index.js generate, so that we know what's getting used where.

As a side effect, this allows for scrapes to generate raw data files (in dist-raw), and gives the option to generate the reports using these raw data files.  These are two new cli methods, `raw:scrape` and `raw:report`, so that we don't get confused with these and the existing `yarn start`.

This then adds two another command, `yarn raw:reportcombined` to combine the data from `yarn raw:scrape` (files stored in `dist-raw`) with the data exported from Li using `tools/gen-raw-files` in that project -- see [Li PR 89](https://github.com/covidatlas/li/pull/89/files#diff-3f7f0275f581ac067f87534341aeb4c5) .  When that runs, it loads the sources and data from Li at the top of the report:

```
$ node src/shared/cli/rawcombinedreport.js -d 2020-04-28
Restoring locations and report from prior saved raw files.
Getting Li data
Loading /raw-li-sources-2020-04-28.json
Loading /raw-li-locations-2020-04-28.json
Added 14 sources and 3110 locations from Li.
... etc, report gen continues as usual.
```

The Li sources are given a high priority (100 + old priority).

## Changes

* Major change: breaks apart the implicit argument passing of src/shared/index.js generate, so that we know what's getting used where.
* Changed the function calls to standardize how the "report summary" is built
* Add `raw:scrape`, `raw:report`, `raw:combinedreport` yarn scripts and js files
* Changed raw data output to include all necessary parts so that reports generated from dist-raw have parity with reports generated the usual way.

The diffs look scattered ... the individual commits might be easier to read, to understand the progression of the change.

## Things noted during report generation

Samples of things from console log output, in no particular order:

```
* Crosscheck failed for fips:01123, iso2:us-al, iso1:us: ../../../../li/src/shared/sources/nyt/index.js (99.31372549019608) has different data than US/nyt-counties.js (-0.6862745098039216)

* Crosscheck failed for iso2:us-wy, iso1:us: ../../../../li/src/shared/sources/nyt/index.js (99.27450980392157) has different data than US/nyt-counties.js (-0.7647058823529411)

[...]

PR #995, opened 2020-05-12
Generate timeseries with cds and li

jzohrab:
Builds off of #968 - "generate reports with cds and li".  

## Summary

This branch changes the report generation so that we have an option to include Li raw data in the report pipeline.  The branch assumes that Li is checked out in the same directory as this branch, and that master is checked out in that repo.

This branch changes the routines that generate reports, but the interface is still the same:

* `yarn start --date x` generates reports.
* `yarn timeseries --date x --endDate y` generates timeseries.
* `yarn timeseries --date x` generates timeseries.

I have run each of these for this branch and for master, and will provide regression results (using `tools/compare-report-dirs.js`)

This change adds a `--addli` option to both `start` and `timeseries`.  If this option is included, then at the start of the report generation a script in Li is called (via `tools/generate-li-dist-raw-files.js`) to generate raw files to `dist-raw`.  The CDS code then incorporates those files into the report generation pipeline.

## Changes

* Refactored report generation, to understand where data needed to be injected
* Refactored the timeseries code thinking that was necessary to make the change I wanted, but eventually realized that the only commit that I needed was really https://github.com/covidatlas/coronadatascraper/commit/5bde483b046748ca4b7f7b278bb170a1abfe7cd8.  !!!  The refactoring is still good, and potentially makes _some_ of the code more portable/available for use in Li.

## Verification / Regression testing

TBD


jzohrab:
Closing this *giant* PR -- we're generating reports with Li now, and will close this project eventually.  This is a big and involved change which had value at one point, but no longer.  Closing! z

PR #998, opened 2020-05-12
Update TW URL description

ndom91:
Hi Guys,

So we're using this as a datasource for our serverless API ([CoronaAPI/coronaAWS](https://github.com/coronaapi/coronaaws)).

When executing the scraper and importing the resulting data into DynamoDB however, the import fails due to the empty string in this specific crawler.

Hope this is alright!


jzohrab:
Perfectly good, thanks for the patch!

PR #1005, opened 2020-05-15
Change incorrect DeprecatedError to NotImplementedError

jzohrab:
I believe this are incorrectly marked, given the comments.  They're not deprecated, they need to be updated.


jzohrab:
(This fails CI b/c the scrapers were touched, and the NotImplementedError is thrown)

jzohrab:
Merging.  I think I am correct, and if I'm not, we can undo this. :-)

PR #1010, opened 2020-05-18
Show deaths rate in addition to confirmed cases rate

gena:
This should add an option to show the death rate per population in addition to the confirmed infected cases, as shown here:

pr-cases-deaths

### DIFF
--- a/site/lib/color.js
+++ b/site/lib/color.js
@@ -153,7 +153,7 @@ export function createLegend(min, max, legendType) {
 
   const heading = document.createElement('span');
   heading.className = 'spectrum-Heading spectrum-Heading--XXXS';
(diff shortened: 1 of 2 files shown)


lazd:
Can you make it a `<select>` instead of a checkbox?

gena:
> Can you make it a `<select>` instead of a checkbox?
Done
image

Maybe it's better to name this _Infected_ instead of _Cases_. 

This can be probably extended to other variables, like _Recovered_, _Admitted to ICU_, etc.

jzohrab:
@gena - that's super!  We've been moving some code around, so I'm not sure of the destination for the PR.  @lazd - is this PR in the right place?

PR #1022, opened 2020-05-20
Replace San Francisco scraper with official API

### LINKED ISSUE #1011: Add scraper for San Francisco
## Location name
San Francisco County, CA, USA

## Source URL
https://data.sfgov.org/COVID-19/COVID-19-Cases-Summarized-by-Date-Transmission-and/tvq9-ec9w/
https://data.sfgov.org/COVID-19/COVID-19-Hospitalizations/nxjg-bhem

## Notes/comments
Like the Santa Clara County Public Health Department (#965), the San Francisco Department of Public Health is now reporting a full time series by date of sample collection rather than reporting date. This means many days’ counts can be retroactively updated on any given day. They’re reporting historical data for all the counts, not just confirmed cases, though they only seem to be providing hospitalization counts up to a certain number of weeks back.

There’s [already a scraper for San Francisco](https://github.com/covidatlas/coronadatascraper/blob/master/src/shared/scrapers/US/CA/san-francisco-county.js), but it scrapes the main dashboard in a way that’s more fragile and doesn’t reflect these retroactive updates.

Unlike Santa Clara County, this source is quite easy to work with. They provide a CSV for download or a REST API to get the table as JSON. As a point of reference, [this Bash script](https://gist.github.com/1ec5/88c9670dcd108d1493a957fb6ac8d819) updates [a table at Wikimedia Commons](https://commons.wikimedia.org/wiki/Data:COVID-19_cases_in_San_Francisco.tab) (used on [this Wikipedia article](https://en.wikipedia.org/wiki/COVID-19_pandemic_in_the_San_Francisco_Bay_Area#Statistics)) with the latest data using the API. I’m not suggesting to rely on Wikimedia Commons or Wikipedia, but perhaps some of the `jq` invocations could be ported to JavaScript.

1ec5:
## Summary

Replaced the scraper for the City and County of San Francisco with a new implementation that joins and pivots three datasets provided by the city’s DataSF portal in JSON format through an official API:

* [COVID-19 Cases Summarized by Date, Transmission and Case Disposition](https://data.sfgov.org/COVID-19/COVID-19-Cases-Summarized-by-Date-Transmission-and/tvq9-ec9w/)
* [COVID-19 Hospitalizations](https://data.sfgov.org/COVID-19/COVID-19-Hospitalizations/nxjg-bhem)
* [COVID-19 Tests](https://data.sfgov.org/COVID-19/COVID-19-Tests/nfpa-mg4g)

Previously, this project scraped the Department of Public Health’s COVID-19 landing page for the current statistics. The page didn’t provide any historical data, so this project would cache past days’ values of the current case total, and clients would tend to treat those values as the actual infection totals as of those days.

Now the project pulls in the full time series as revised by the department each day, which is optimal for charting infection rates. It also adds hospitalization and test counts.

Fixes #1011.

## Changes

This implementation pivots each dataset on the date field, summing case counts and distinguishing between case count and death toll for the first dataset.

### DIFF
--- a/src/shared/scrapers/US/CA/san-francisco-county.js
+++ b/src/shared/scrapers/US/CA/san-francisco-county.js
@@ -1,5 +1,5 @@
 import * as fetch from '../../../lib/fetch/index.js';
-import * as parse from '../../../lib/parse.js';
+import datetime from '../../../lib/datetime/index.js';
 import maintainers from '../../../lib/maintainers.js';
 
 // Set county to this if you only have state data, but this isn't the entire state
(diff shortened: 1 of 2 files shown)


jzohrab:
Super work @1ec5, thank you!  I'll look into this one today.  Cheers!

jzohrab:
Today got away from me, I'll schedule time to look at this tomorrow.

1ec5:
There’s something wrong with this scraper, but I can’t put my finger on it. It’s coming up with inaccurate case totals because of what seems to be a stale cache of the API response. For example, the cached copy of [tvq9-ec9w.json](https://data.sfgov.org/resource/tvq9-ec9w.json) still has an entry for a new confirmed case on March 7, but the current dataset no longer includes any new cases for March 7. (That case probably got moved to a different date.)

src/shared/scrapers/US/CA/san-francisco-county.js
```diff
@@ -9,26 +9,99 @@ const scraper = {
...
+      tests.reduceRight((acc, cur) => {
+        const date = datetime.getYYYYMMDD(cur.result_date);
+        acc += +cur.tests;
+        if (!(date in timeSeries)) {
+          timeSeries[date] = {};
+        }
+        timeSeries[date].tested = acc;
+        return acc;
+      }, 0);
+
```
jzohrab, line 82: The undocumented standard :-) for scraping in this project appears to be checking date bounds at the beginning, and filtering data before scraping.  At least, that's what it should be.  Your method works too, so it's good and no need to change it, just a note for the future!

jzohrab:
Hm, getting inaccurate numbers, I think.  After running this, I get 

```
   - 0 cities
   - 0 states
   - 1 counties
   - 0 countries
   - 968 cases
   - 35266 tested
   - 0 recovered
   - 0 deaths
   - 0 active
```

0 deaths, but https://data.sfgov.org/resource/tvq9-ec9w.json shows deaths.  Looking into it.

jzohrab:
A few issues:

* the incoming data isn't sorted by date, so the running totals aren't being done in the right order.
* "reduceRight" takes things off of the end of the array.  If the data is in date order, that would result in the running totals being backwards (starting off at 0 as at latest date, and then increasing).  "reduce" is correct.
* not every date is present in the array, so we end up with data like this, eg:

```
┌────────────┬───────┬────────┐
│  (index)   │ cases │ deaths │
├────────────┼───────┼────────┤
│ 2020-03-05 │   2   │        │
│ 2020-03-06 │   6   │        │
│ 2020-03-08 │  11   │        │
...
│ 2020-03-16 │  37   │        │
│ 2020-03-17 │  48   │   1    │
│ 2020-03-18 │  62   │        │
...
│ 2020-03-23 │  165  │        │
│ 2020-03-24 │  194  │   2    │
│ 2020-03-25 │  243  │        │
│ 2020-03-26 │  279  │   5    │
│ 2020-03-27 │  315  │   7    │
│ 2020-03-28 │  344  │        │
│ 2020-03-29 │  381  │        │
│ 2020-03-30 │  415  │   8    │
│ 2020-03-31 │  442  │   9    │
│ 2020-04-01 │  481  │   12   │
│ 2020-04-02 │  542  │        │
```

So that would look like deaths = '' on 04-02, but it should be 12.

Working on some changes, which I'll push here before merging.  Thanks!

src/shared/scrapers/US/CA/san-francisco-county.js
1ec5: The data source only provides the number of new cases on each day, whereas this project seems to expect a cumulative number on each day. Calculating the cumulative number unfortunately requires reducing from the end of the array, but we could have the `reduceRight` call stop at the requested date if performance is a concern.

1ec5:
> "reduceRight" takes things off of the end of the array. If the data is in date order, that would result in the running totals being backwards (starting off at 0 as at latest date, and then increasing). "reduce" is correct.

When I had originally fetched the JSON files, they were coming in in reverse chronological order. But you’re right, it’s totally unsorted now. They must’ve created the dataset with some other data that was already in reverse chronological order but then continued to add records without sorting them. Thanks for looking into this issue!

jzohrab:
Ah interesting about the sorting changes.  Cheers, still working on it!

jzohrab:
I opened a PR to this branch: https://github.com/1ec5/coronadatascraper/pull/2

As you said the source data changed, I think what I suggest is foolproof. :-)

jzohrab:
Hi @1ec5 - are you good with my PR, can you merge it, and then we merge this one?

jzohrab:
Closed, replaced by https://github.com/covidatlas/coronadatascraper/pull/1044.

:tada:

PR #1027, opened 2020-05-22
Add testing counts to US/MO

davidcardonAH:
## Summary

Adds testing county-level test counts to Missouri state data

Previously, testing counts were not collected at the county level

This change adds that by pulling from 

## Changes

- Add a new utility function for querying arcGIS data sources, handles pagination.
- Add a function to scrapers/US/MO.index.js that enhances the existing case and death count data for the county with testing counts

### DIFF
--- a/docs/sources.md
+++ b/docs/sources.md
@@ -431,7 +431,7 @@ You should run your source with the crawler by running `yarn start -l "<path to 
 
 The path to scraper should be the relative path under
 `src/shared/scrapers`; e.g., the scraper for Montana, USA would be
-"US/MO").
+"US/MT").
 
 After the crawler has finished running, look at how many counties, states, and countries were
 scraped. Also look for missing location or population information. Finally, look at the output located in the `dist` directory. `data.json` contains all the information
--- a/src/shared/lib/fetch/index.js
+++ b/src/shared/lib/fetch/index.js
@@ -293,3 +293,46 @@ export const getArcGISCSVURL = async function(scraper, serverNumber, dashboardId
   const { orgId } = dashboardManifest;
   return getArcGISCSVURLFromOrgId(scraper, serverNumber, orgId, layerName);
 };
+
+/**
+ * Get the full JSON response from arcGIS.  The query URL format is parameterized like this:
+ *  https://services{serverNumber:int}.arcgis.com/{orgId:string,alphanumeric}/arcgis/rest/services/{layerName:string}/FeatureServer/0/query
+ *
+ * You can browse the repository of layers within an orgId by going to URL: https://services{serverNumber}.arcgis.com/{orgId}/
+ * Once there, you can examine the various datasets / layers available for that organization.
+ *
+ * This function accumulates multiple pages of responses into the return object's "features" array.
+ *
+ * @param {*} scraper the scraper object
+ * @param {*} serverNumber the arcGIS server number (see format above)
+ * @param {*} orgId parameter for the arcGIS server request (see format above)
+ * @param {*} layerName parameter for the arcGIS server request (see format above)
+ * @param {*} options the request options, including the query where clause and any aggregation options
+ * @param {*} date the date of the request--populate this if caching is desired, otherwise set to false
+ * @param {*} cacheKey the key / namespace to use in generating the cache id
+ */
+export const queryArcGISJSON = async function(
+  scraper,
+  serverNumber,
+  orgId,
+  layerName,
+  options,
+  date,
+  cacheKey = 'default'
+) {
+  const url = `https://services${serverNumber}.arcgis.com/${orgId}/arcgis/rest/services/${layerName}/FeatureServer/0/query`;
+  let result = await json(scraper, url, cacheKey, date, options);
+  const rval = result;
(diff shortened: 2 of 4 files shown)


src/shared/lib/fetch/index.js
```diff
@@ -293,3 +293,46 @@ export const getArcGISCSVURL = async function(scraper, serverNumber, dashboardId
...
+ *
+ * @param {*} scraper the scraper object
+ * @param {*} serverNumber the arcGIS server number (see format above)
+ * @param {*} orgId parameter for the arcGIS server request (see format above)
+ * @param {*} layerName parameter for the arcGIS server request (see format above)
+ * @param {*} options the request options, including the query where clause and any aggregation options
+ * @param {*} date the date of the request--populate this if caching is desired, otherwise set to false
+ * @param {*} cacheKey the key / namespace to use in generating the cache id
+ */
+export const queryArcGISJSON = async function(
```
davidcardonAH, line 314: This is a first stab at a way to get paged data out of arcGIS. I've only tested this repeatedly on the Missouri test case data set.

src/shared/scrapers/US/MO/index.js
```diff
@@ -168,6 +194,59 @@ const scraper = {
...
+  // Missouri stores test result data separately from death and case counts, which are reported on / updated only
+  // daily as cumulative counts. This function downloads the cumulative testing counts for current scrape
+  // date.
+  async _applyTestingCounts(counties) {
+    const captureDate = new Date(datetime.scrapeDate() || getYYYYMMDD());
+
+    const queryParams = {
+      args: {
+        where: `test_date <= DATE '${getYYYYMMDD(captureDate)}'`,
+        groupByFieldsForStatistics: 'county,result',
```
davidcardonAH, line 206: This query relies on arcGIS to do the aggregation, at least to the county and test result (Positive, Negative, Indeterminate) level. Incidentally, the positive case counts in this data set differs slightly from the historically, snapshotted data previously collected by this scraper (these positive case counts are slightly higher). It's possible that the snapshotted counts are lower due to reporting lag, but I'm curious which of the counts would be preferred in the report.

src/shared/scrapers/US/MO/index.js
```diff
@@ -198,6 +277,7 @@ const scraper = {
           }
         }
       });
+      await this._applyTestingCounts(counties);
```
davidcardonAH, line 280: Since we have historical testing data for all three scraper time periods (0, '2020-02-22' and '2020-03-30'), the test counts get applied to each one.

tests/integration/scrapers/scrapers-new-test.js
```diff
@@ -34,6 +34,7 @@ if (files) {
           const hasErrors = schema.schemaHasErrors(scraper.default, schema.schemas.scraperSchema);
           t.notOk(hasErrors, 'Scraper had no errors');
         } catch (err) {
+          console.log(err);
```
davidcardonAH, line 37: I noticed that on test failures the full stack trace of the error wasn't visible. I thought having it visible at this point in the code would be a good addition, since it definitely helped me in debugging.

docs/sources.md
```diff
@@ -431,7 +431,7 @@ You should run your source with the crawler by running `yarn start -l "<path to
 
 The path to scraper should be the relative path under
 `src/shared/scrapers`; e.g., the scraper for Montana, USA would be
-"US/MO").
+"US/MT").
```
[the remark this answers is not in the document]
jzohrab: Thank you!  Perhaps a Canadian wrote that doc.

jzohrab:
Hi @davidcardonAH - apologies for the _extremely_ long delay in getting to this PR.  I've been working on porting things to our new version of this project, and had quite a few things to keep in the air until that was done.  I will move this way up on my list of priorities to port to the new project.

Thanks again, it's disheartening to do good work only to have it languish unused.  Sorry about that!  Regards, jz

jzohrab:
I'll change this while porting to Li.  The code for submitting the query using the form params is great, thanks.

URL: https://services6.arcgis.com/Bd4MACzvEukoZ9mR/ArcGIS/rest/services/Daily_COVID19_Testing_Report_for_OPI/FeatureServer/0/query

Sample data in the form:
image

Note that this result is paginated.

One change I'll make when porting: Li has the concept of "multivalent data", meaning that data from several sources can be pulled into the final reports.  So, rather than us going through the gymnastics of shoving the data into the existing us/mo/index.js source, we can create a new source that just gets this data.  That will simplify the change.  Second, since it's separate, we can actually pull in the full timeseries at once ... we'll have to calculate the running sums, but that should be relatively easy.

There are a lot of changes to the submitted code due to the project changes, but your work provides a great starting point.

jzohrab:
I've added a new method for arcgis pagination handling in Li, and can soon port the rest of this over to a new source there.

jzohrab:
I've merged https://github.com/covidatlas/li/pull/385 into Li master, and will deploy it to production shortly.  This project isn't merging PRs anymore.

Thanks @davidcardonAH , this was great work that I borrowed/stole to improve the Li code.  We now have arcgis pagination with query-style params as you've used here.  Cheers! jz

Issue #1000, opened 2020-05-13
Add scraper for Orleans Parish, Louisiana (i.e. New Orleans)


mikelehen:
## Location name
Orleans Parish, LA, USA

## Source URL
http://nolagis.maps.arcgis.com/apps/opsdashboard/index.html#/7944c56c0b914072bbd9a8f4cceb0ee1
linked from (county website)
https://ready.nola.gov/incident/coronavirus/safe-reopening/ (under "Covid-19 dashboard")

## Notes/comments
Currently you only have cases / deaths, but the dashboard provides additional information including "total tests" which is what I specifically want.

mikelehen:
For reference, it looks like the underlying dataset can be fetched as https://gis.nola.gov/arcgis/rest/services/apps/LDH_Data/MapServer/0/query?f=json&where=Date%3Etimestamp%20%272020-03-18%2003%3A59%3A59%27&returnGeometry=false&spatialRel=esriSpatialRelIntersects&outFields=*&orderByFields=Date%20asc&outSR=102100&resultOffset=0&resultRecordCount=1000

sglyon:
I would like to contribute here, specifically to get information on hospital and icu capacity and current load.

From [the CDS docs](https://github.com/covidatlas/coronadatascraper/blame/master/docs/data_fields.md#L48), it seems that the variable `icu` tracks the cumulative number of covid patients that have been admitted to the icu. 

The variables available on the dashboard in the original link are:

```
['OBJECTID',
 'Date',
 'NO_Cases',
 'NO_Deaths',
 'NO_State_Tests',
 'NO_Commercial_Tests',
 'NO_Total_Tests',
 'NO_Tests_per_100k',
 'NO_Cases_per_100k',
 'NO_Deaths_per_100k',
 'NO_Positive_Rate',
 'NO_Deaths_per_Positive',
 'LA_Cases',
 'LA_Deaths',
 'LA_State_Tests',
 'LA_Commercial_Tests',
 'LA_Total_Tests',
 'LA_Tests_per_100k',
 'LA_Cases_per_100k',
 'LA_Deaths_per_100k',
 'LA_Positive_Rate',
 'LA_Deaths_per_Positive',
 'LA_Patients_Hospitalized',
 'LA_Patients_Ventilated',
 'LA_COVID_Hospitalizations',
 'LA_COVID_Hospitalizations_Delta',
 'R1_Ventilators_In_Use',
 'R1_ICU_Beds_In_Use',
[...]

Issue #940, opened 2020-05-02
Support ICU data


lazd:
## Description
Cumulative field, `icu`, with total number of patients that were admitted to the ICU.

This will need a corollary field, `icu_discharged` or `icu_current` or something, which we can deal with later.

## Why do you need this feature or component?

Tracking severity of cases, when the data is available.

PR #1030, opened 2020-05-22
Add support for icu_current and hospitalized_current

lazd:
## Summary

Add support for icu_current and hospitalized_current

Related to #940 

Related to #1000

## Changes

* Validation for `hospitalized` and `discharged` -- will catch locations that were incorrectly noting hospitalized data as cumulative when it should have been current
* Fixup Washoe (was incorrectly noting hospitalized data as cumulative when it should have been current)
* Add `icu_current` and `hospitalized_current` to Mercury News scraper
* Add docs


jzohrab:
LGTM!

PR #1003, opened 2020-05-13
Graph

webguyty:
## Summary

Basic graph in early stages showing an overview of all the dates displaying data in linear or logarithmic function. 

Bar chart showing daily cases or deaths. UI for development.

Have not styled yet.

PR #1069, opened 2020-07-26
graph made with d3

webguyty:


jzohrab:
Replacement for #1003.

